PR #9, opened 2020-01-16
mergify: add config yaml

SteVwonder:
@grondo: any idea how to add a list to the `approved-reviews-by` condition?  I attempted to use yaml arrays, but the mergify validation (`curl -F 'data=@.mergify.yml' https://gh.mergify.io/validate`) rejects it.

Both `- approved-reviews-by=[@flux-framework/core,@flux-framework/docs]` and

```yaml
- approved-reviews-by=
  - @flux-framework/core 
  - @flux-framework/docs
```

are rejected.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,17 @@
+pull_request_rules:
+  - name: rebase and merge when passing all checks
+    conditions:
+      - base=master
+      - status-success=continuous-integration/travis-ci/pr
+      - status-success="validate commits"
+      - label="merge-when-passing"
+      - label!="work-in-progress"
+      - "approved-reviews-by=@flux-framework/docs"
+      - "#approved-reviews-by>0"
+      - "#changes-requested-reviews-by=0"
+      - -title~=^\[*[Ww][Ii][Pp]
+    actions:
+      merge:
+        method: merge
+        strict: smart
+        strict_method: rebase
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,12 @@
+on: pull_request
+jobs:
+  check-pr:
+    name: validate commits
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 0
+    - run: git fetch origin master
+    - uses: flux-framework/pr-validator@master


grondo:
This seems like it may be a bug in mergify.io. I think the values in conditions are sent to the mergify engine as is (i.e. are not themselves yaml keys). The [documentation](https://github.com/Mergifyio/mergify-engine) does state that `approved-reviews-by` is an "array of string", but I think the teams `@` character is confusing the array parser.

I found [the simulator](https://dashboard.mergify.io/installation/661098/simulator?repositoryName=flux-core) useful, you can even run it against an open PR. I was able to make a list of users "work" with a simple comma-separated list, but if I added an @ character, it failed to parse.

The code seems to be here:

 https://github.com/Mergifyio/mergify-engine

maybe you can figure out what is going on.

SteVwonder:
Thanks @grondo!  Poking around in the rules parser, it looks like they only allow a single team or username.  Not an array.  Relevant source line: https://github.com/Mergifyio/mergify-engine/blob/4d8f774d801355eb5aca6a8791d3b16367cdfab8/mergify_engine/rules/parser.py#L21

I opened an issue upstream to confirm: https://github.com/Mergifyio/mergify-engine/issues/656

Maybe we merge this PR as-is and get the multiple teams working later.

.mergify.yml
```diff
@@ -0,0 +1,17 @@
+pull_request_rules:
+  - name: rebase and merge when passing all checks
+    conditions:
+      - base=master
+      - status-success=continuous-integration/travis-ci/pr
+      - status-success="validate commits"
```
grondo: Do we plan on enabling the pr-validator for this repo? If so maybe that can go in this PR as well. If not then the `"validate commits"` status can be removed from the mergify config.

SteVwonder: Great suggestion!  Added the action yml in this PR.

SteVwonder:
I heard back on the upstream issue.  Looks like adding the second team requires a duplicate rule.  I'm waiting to hear back if the duplicate rule will somehow allow the action to be performed twice.

SteVwonder:
It looks like duplicating the rule does cause the action to run twice.  So I think we stick with a single rule that uses `@flux-framework/docs`.  Just pushed that change.

grondo:
Thanks @SteVwonder, merging.

Issue #14, opened 2020-01-16
quickstart: update `flux module` and `flux job list` usage


SteVwonder:
After the most recent release (i.e., in current master), the flux module command lost the --rank argument, and the flux job list command moved from being a "porcelain" command to being a "plumbing" command. It has been replaced by "flux jobs".

The quickstart.rst should be updated to reflect these changes.

cmoussa1:
This issue should be fixed by PR #20, so I'll go ahead and close this.

grondo:
Note that `flux job list` default output changed again :joy: 

The output can perhaps be updated once we generalize and unify usernames in all examples in the docs.

cmoussa1:
Oops! Thanks for clarifying @grondo 🤣 maybe I should re-open this for now? Do you think those changes could be made in @SteVwonder's current PR?

grondo:
Might as well do it in the fix for #26

PR #20, opened 2020-02-21
quickstart: Update 'flux module', 'flux job list' usage

cmoussa1:
#### Background

From issue #14: 

> After the most recent release (i.e., in current master), the flux module command lost the --rank argument, and the flux job list command moved from being a "porcelain" command to being a "plumbing" command. It has been replaced by "flux jobs".

#### Fix

This PR addresses #14 and updates the following commands to reflect the changes in [flux-framework/flux-core](https://github.com/flux-framework/flux-core), specifically the drop of `--rank` from `flux module list` and the introduction of `flux jobs`, the new job listing tool.

### DIFF
--- a/quickstart.rst
+++ b/quickstart.rst
@@ -157,22 +157,21 @@ Services within a Flux session may be implemented by modules loaded in the ``flu
 
 .. code-block:: console
 
-  $ flux module list --rank=all
-  Module                   Size Digest  Idle  S   Nodeset Service
-  job-ingest            1143808 AF0B59E    4  S     [0-2]
-  cron                  1150080 4734B87    0  S         0
-  connector-local       1058616 30B79F6    0  R     [0-2]
-  resource             15996328 C5E139A    3  S         0
-  qmanager               841040 D40C2E0    3  S         0 sched
-  userdb                1066296 56B5D15    4  S         0
-  content-sqlite        1074088 7098AD8    0  S         0 content-backing
-  job-manager           1200296 D97A9A2    3  S         0
-  kvs                   1485416 23988F3    0  S     [0-2]
-  kvs-watch             1231456 4F4618B    0  S     [0-2]
-  job-info              1178896 CEB665F    4  S     [0-2]
-  barrier               1067888 8AEA814    4  S     [0-2]
-  aggregator            1085000 D967157    0  S     [0-2]
-  job-exec              1176856 CED941E    4  S         0
+  $ flux module list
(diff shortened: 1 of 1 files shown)


quickstart.rst
```diff
@@ -275,7 +275,7 @@ Here, the allocated ID for the job is immediately echoed to stdout.
 
 .. code-block:: console
 
-  $ flux job list
+  $ flux jobs
     JOBID           STATE   USERID    PRI   T_SUBMIT
     640671547392    R       58985     16    2019-10-22T16:27:02Z
     1045388328960   R       58985     16    2019-10-22T16:27:26Z
```
SteVwonder: The output of the `flux jobs` command also differs from `flux job list`, right?  Do you mind also updating that with the new format?

cmoussa1: Ah! Yes, you're right. That output is different. I went ahead and updated the output, squashed and force-pushed.

cmoussa1: Looking more closely, it looks like the `NAME` field from `flux jobs` gets truncated after a small number of characters. Is there a way to increase the length of that field to accommodate for a longer name?

grondo: The default field width for the name is 10 characters. To increase the width of the field you'll have to override the default format for flux-jobs, e.g.

```console
$ flux mini submit --job-name=reallylongnameforsomereason sleep 300
1041965776896
$ flux jobs
             JOBID USER     NAME       STATE    NTASKS NNODES  RUNTIME RANKS
     1041965776896 grondo   reallylong RUN           1      1   1.588s 0
$ flux jobs --format="{name:<16}"
NAME            
reallylongnameforsomereason
```

However, that may be out-of-scope for a Quickstart document.

quickstart.rst
```diff
@@ -157,7 +157,7 @@ Services within a Flux session may be implemented by modules loaded in the ``flu
 
 .. code-block:: console
 
-  $ flux module list --rank=all
+  $ flux module list
   Module                   Size Digest  Idle  S   Nodeset Service
   job-ingest            1143808 AF0B59E    4  S     [0-2]
```
grondo: The output of `flux module list` also needs to be updated. It no longer includes `Nodeset` since it only targets a single rank.

cmoussa1: Thanks for catching that @grondo! I went ahead and updated the output and force-pushed.

cmoussa1:
Thanks! 🎉

PR #21, opened 2020-02-22
[WIP] RFC integration

gonsie:
Add contents from rfc repo.

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -11,12 +11,35 @@
 Welcome to Flux's documentation!
 ================================
 
-.. toctree::
-   :maxdepth: 3
-   :caption: Contents
+Flux is a flexible framework for resource management, built for your site.
+The framework consists of a suite of projects, tools, and libraries which may be used to build site-custom resource managers for High Performance Computing centers.
 
-   quickstart
+.. toctree::
+   :maxdepth: 2
+   :caption: Overview
+
    contributing
+   quickstart
+
+.. toctree::
+   :maxdepth: 1
+   :caption: User Relevant RFCs
+
+   rfc/spec_1
+   rfc/spec_2
+   rfc/spec_7
+   rfc/spec_9
+
+.. toctree::
+   :maxdepth: 1
+   :caption: API Reference
+
+   rfc/README.md
+
+.. toctree::
+   :hidden:
+
+   rfc/index
 
 Indices and tables
 ==================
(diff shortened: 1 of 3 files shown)


SteVwonder:
Thanks @gonsie for putting this together.  What else need to happen before this can be considered no longer a WIP?

Few suggestions to things to add (we can make them a separate PR if you want):

- Adding a requirements.txt for the python dependencies to be installed via pip
- Adding a line to the readme that is easily copy-and-pasteable for installing the packages via apt:

```
sudo apt install python3-sphinx python3-recommonmark python3-sphinxcontrib.spelling python3-sphinx-rtd-theme
```

- Adding a line about how to properly clone this repo now that it has submodules:

```
git clone --recurse-submodules git@github.com:flux-framework/flux-docs.git
```

---

Also worth noting that the `python3-recommonmark` in Ubuntu 19.10 (i.e., not a LTS) is 0.4.0, which suffers from [this bug](https://github.com/readthedocs/recommonmark/issues/119), which breaks recommonmark with the sphinx that is also packaged with 19.10.  Not sure if using the system packages will work for the LTSes (i.e., 18.04 and 20.04).  So maybe we don't want to advertise using the system package manager in our README.

index.rst
```diff
@@ -11,12 +11,35 @@
...
+   :maxdepth: 2
+   :caption: Overview
 
-   quickstart
    contributing
+   quickstart
+
+.. toctree::
+   :maxdepth: 1
+   :caption: User Relevant RFCs
```
SteVwonder: Looking at this in the context of the rendered page, I think these RFCs are more relevant to developers than users:
```suggestion
   :caption: Contributor Relevant RFCs
```

index.rst
```diff
@@ -11,12 +11,35 @@
...
+   rfc/spec_1
+   rfc/spec_2
+   rfc/spec_7
+   rfc/spec_9
+
+.. toctree::
+   :maxdepth: 1
+   :caption: API Reference
+
+   rfc/README.md
```
SteVwonder: I think we chatted about this at the flux doc-a-thon, but I forget what the outcome was.  Is it possible to change the way the name here (and in the sidebar) renders from just `rfc` to `Flux RFC Index`?

gonsie:
We’ve decided to move away from git submodules.

PR #27, opened 2020-03-02
quickstart.rst: update flux-sched build instructions

grondo:
Update flux-sched instructions now that `PKG_CONFIG_PATH` and `PYTHON_VERSION` are not required to build against side-installed flux-core.


grondo:
Thanks!

PR #30, opened 2020-03-07
:heavy_minus_sign: removed flux-core as git submodule

gonsie:
We will be switching to using [RTD Subprojects](https://docs.readthedocs.io/en/stable/subprojects.html) instead of git submodules.


SteVwonder:
@mergifyio refresh

PR #52, opened 2020-07-27
faqs: Fix typo

chu11:


faqs.rst
```diff
@@ -68,7 +68,7 @@ An example command to set the ``rundir`` could look like:
 How do I mimic Slurm's jobstep semantics ?
 ------------------------------------------
 
-``flux mini submit`` ting a script with ``flux mini run`` s will not result in Slurm-style jobsteps unless you stick ``flux start`` in front of the jobscript in ``flux mini submit`` .
+``flux mini submit`` ing a script with ``flux mini run`` s will not result in Slurm-style jobsteps unless you stick ``flux start`` in front of the jobscript in ``flux mini submit`` .
```
grondo, line 71: actually isn't two `t`s correct for submitting? Perhaps use `-ing` (i.e. prefix with dash), or avoid attempting to append `-ing` to our commands in the docs, since doing this with multi-word commands is going to get confusing. This sentence could be reworded:
```suggestion
Using ``flux mini submit`` to submit a script containing multiple ``flux mini run`` invocations will not result in Slurm-style job steps unless the job script is prefixed with ``flux start``
```

Side note: Job steps are not usually called "jobsteps" and  a job script is not usually called a "jobscript", so we should avoid those in our docs.

chu11: ohh i get it ... completing the word "submit-ting".   Yeah, I was thinking more generally `-ing` across a multi-word phrase.

I'll just go with your re-word above.  Which, given its the entirety of this patch, I'll make you the author :-)

chu11:
re-pushed with @grondo's suggestion above.  Suppose someone else should push the button since we've both now touched this PR :-)  Although it is documentation, an accepted exception.

grondo:
Hm, sphinx is complaining now:

```
##[warning]Title overline too short.
##[warning]Title overline too short.
[sphinx-action] Build failed with 2 warnings
```

chu11:
i see what I did, repushed

faqs.rst
SteVwonder: > Job steps are not usually called "jobsteps" and a job script is not usually called a "jobscript", so we should avoid those in our docs.

We should pull those out of the spelling dictionary then too

grondo:
Sounds good, I'll do that. I'll also add a fix to have `make check` emit the misspelled words.

grondo:
Ok, removed `jobscript` and `jobstep` from spelling dictionary. Also added a commit to emit misspelled words on `make check` failure.

PR #55, opened 2020-08-04
Add parallel debugging section 

dongahn:
- Add doc on how to debug parallel jobs using TotalView
Introduce basic TotalView commands
for launch and attach mode

- Add ways to improve Flux's own internal exec
events within TotalView.

- Add known issues section to document
current limitations.

- Increase flux-core version number to 0.19.0

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -33,9 +33,6 @@ copyright = '''Copyright 2014 Lawrence Livermore National Security, LLC and Flux
 SPDX-License-Identifier: LGPL-3.0'''
 author = 'This page is maintained by the Flux community.'
 
-# The full version, including alpha/beta/rc tags
-release = '0.13.0'
-
 # -- RTD configuration -------------------------------------------------------
 
 # on_rtd is whether we are on readthedocs.org, this line of code grabbed from docs.readthedocs.org
--- a/debugging.rst
+++ b/debugging.rst
@@ -1,0 +1,99 @@
+.. _debugging:
+
+==============
+Debugging Jobs 
+==============
+
+Flux supports parallel debuggers such as Rogue Wave Software (RWS)'s
+`TotalView parallel debugger <https://totalview.io>`_.
+
+----------------------------------
+Parallel Debugging using TotalView
+----------------------------------
+
+Debugging your MPI job from the beginning of
+parallel program execution:
+
+.. code-block:: console
+
+  $ totalview --args flux mini run -N 2 -n 2 ./mpi-program
+
+Attaching to an already running job:
+
+.. code-block:: console
+
+  $ flux job attach --debug <JOBID> &
+  $ PID=$!
+  $ totalview -pid ${PID} /proc/${PID}/exe
+
+You can also just type ``totalview`` without ``-pid`` option and
+then use its ``Start a Debugging Session``
+GUI option to select the process of this ``flux-job`` command:
+``A running program (attach)``.
+
+.. note::
+  You use TotalView with the newly invoked ``flux job attach``
+  when your job has been launched via ``flux mini run`` (or ``flux mini submit``).
+  This is because Flux currently does not allow TotalView
+  to attach to your running code if you just attach it
+  to the ``flux mini run`` process that controls your code. For
+  performance reasons, ``flux mini run`` does not generate
+  sufficient debug information when it is invoked with no tool control.
+
+Flux provides ``MPIR_partial_attach_ok`` `[1] <https://www.mpi-forum.org/docs/mpir-specification-10-11-2010.pdf>`_
+so you can attach TotalView to a subset of parallel processes as well, which
+can be handy when you debug a large-scale job. Please refer to
+`TotalView user guide`_ on its subset-attach capability.
+
+.. _TotalView user guide: https://docs.roguewave.com/en/totalview/current/html/
+
+----------------------------------------------
+Better Handling of Flux's Internal Exec Events
+----------------------------------------------
+
+Flux executes a series of subcommands before launching
+your parallel program.
+Thus, a side effect of the above commands is that TotalView
+will stop Flux's own program execution each time
+a new subcommand is ``exec(3)``'ed (e.g., when ``flux`` ``exec``'s
+``flux-mini``) before your parallel program processes are spawned.
+This means you must manually click "Yes" whenever a stop-at-exec
+dialogue window is popped up. You do not need to stop
+it unless you are a Flux developer!
+Instead, you can make TotalView automatically
+continue on these events by building the the following
+exec-handling Tcl code into TotalView::
+
+    catch {dset TV::exec_handling { {^(flux|lrun|srun|jsrun)(<python[^>]*>(<flux-job>)?|<bash>|<jsrun>|<jswrap>)*$ go}} }
+
+This code should either be added to the site-wide ``.tvdrc`` file
+to enable this for all TotalView sessions
+or per-user file (e.g., ``tvdrc`` in the current working directory)
+to enable this only for the user's own sessions.
+For the latter case, you should provide ``-s tvdrc`` option
+to ``totalview``. For example,
+
+.. code-block:: console
+
+  $ totalview -s tvdrc --args flux mini run -N 2 -n 2 ./mpi-program
+
+Notice that it is designed to support not only Flux but also SLURM's
+srun and IBM JSM's jsrun commands. The ``regex`` syntax of
+``exec_handling`` within TotalView can be found in `TotalView user guide`_.
+
+
+------------
+Known Issues
+------------
+Parallel debugging support within Flux is new and we are working with
+tool builders to mature our functionality further. This section documents
+some of known limitations.
+
+1. Depending on how the site configured TotalView,
+the parallel program may not stop at the first instruction
+to allow TotalView to debug them from the beginning of execution.
+For performance reasons, TotalView can be configured
+not to stop and process dynamic libraries and this
+can cause this issue. RWS is actively looking
+at this issue.
+
--- a/index.rst
+++ b/index.rst
@@ -21,6 +21,7 @@ The framework consists of a suite of projects, tools, and libraries which may be
    quickstart
    contributing
    coral
+   debugging
    faqs
 
 
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -461,3 +461,4 @@ distro
 subcomponents
 readthedocs
 repos
+Tcl


conf.py
```diff
@@ -34,7 +34,7 @@
 author = 'This page is maintained by the Flux community.'
 
 # The full version, including alpha/beta/rc tags
-release = '0.13.0'
+release = '0.19.0'
```
SteVwonder, line 37: We should probably delete this line.  We did so in flux-core, and I don't think this is used anywhere.  @gonsie, can you confirm/do you agree?

If we don't delete it, we should set it to `0.1.0` or `0.0.0` since flux-docs will be versioned independently from flux-core and flux-sched.

debugging.rst
```diff
@@ -0,0 +1,84 @@
+.. _debugging:
+
+==============
+Debugging Jobs 
+==============
+
+Flux supports parallel debuggers such as Rogue Wave Software (RWS)'s
+TotalView parallel debugger (https://www.roguewave.com).
```
SteVwonder, line 8: ReST has support for external links.  The syntax is weird:

```suggestion
Flux supports parallel debuggers such as `Rogue Wave Software (RWS)'s <https://www.roguewave.com>`_
TotalView parallel debugger.
```

debugging.rst
```diff
@@ -0,0 +1,84 @@
...
+
+  totalview --args flux mini attach <JOBID>
+
+Note that you should use ``totalview --args flux mini attach`` whether your
+job has been launched via ``flux mini run`` or ``flux mini submit``.
+In other word, Flux currently does not allow TotalView
+to attach to your running code if you just attach it
+to the ``flux mini run`` process that controls your code.
+
+Flux provides ``MPIR_partial_attach_ok`` (https://www.mpi-forum.org/docs/mpir-specification-10-11-2010.pdf)
```
SteVwonder, line 33: I **think** it is possible to combine the verbatim mode (``) with hyperlinks, but I don't want to suggest something without trying it first.  If it isn't easily doable, maybe something like:

```suggestion
Flux provides ``MPIR_partial_attach_ok`` `[1]<https://www.mpi-forum.org/docs/mpir-specification-10-11-2010.pdf>`_
```

debugging.rst
```diff
@@ -0,0 +1,84 @@
...
+so you can attach TotalView to a subset of parallel processes as well, which
+can be handy when you debug a large-scale job. Please refer to
+TotalView user guide on its subset-attach capability: https://docs.roguewave.com/en/totalview/current/html/.
```
SteVwonder, line 36: Since the TotalView user guide is referenced multiple times, you can created a named link rather than an anonymous one (I **think** this name is valid anywhere in this file, maybe even across the entire repo).

```suggestion
`TotalView user guide`_ on its subset-attach capability.

.. _TotalView user guide: https://docs.roguewave.com/en/totalview/current/html/
```

debugging.rst
```diff
@@ -0,0 +1,84 @@
...
+    catch {dset TV::exec_handling { {^(flux|lrun|srun|jsrun)(<python[^>]*>(<flux-job>)?|<bash>|<jsrun>|<jswrap>)*$ go}} }
+
+This code should either be added to the site-wide ``.tvdrc`` file
+to enable this for all TotalView sessions
+or per-user file (e.g., ``tvdrc`` in the current working directory)
+to enable this only for the user's own sessions.
+Notice that it is designed to support not only Flux but also SLURM's
+srun and IBM JSM's jsrun commands. The ``regex`` syntax of
+``exec_handling`` within TotalView can be found in the TotalView reference
+guide: e.g., https://docs.roguewave.com/en/totalview/current/html/.
```
SteVwonder, line 67: Assuming the named reference suggested above is created:

```suggestion
``exec_handling`` within TotalView can be found in the `TotalView user
guide`_.
```

debugging.rst
```diff
@@ -0,0 +1,84 @@
...
+----------------------------------
+Parallel Debugging using TotalView
+----------------------------------
+
+Debugging your MPI job from the beginning of
+parallel program execution:
+
+.. code-block:: sh
+
+  totalview --args flux mini run -N 2 -n 2 ./mpi-program
```
SteVwonder, line 19: Nit pick.  ReST has a `console` code-block that supports highlighting of the terminal prompt, command, and output.  Since this is just a single line, it doesn't matter all that much.  Mainly an FYI.

```suggestion
.. code-block:: console

  $ totalview --args flux mini run -N 2 -n 2 ./mpi-program
```

debugging.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Attaching to an already running job:
+
+.. code-block:: sh
+
+  totalview --args flux mini attach <JOBID>
+
+Note that you should use ``totalview --args flux mini attach`` whether your
+job has been launched via ``flux mini run`` or ``flux mini submit``.
+In other word, Flux currently does not allow TotalView
```
SteVwonder, line 29:
```suggestion
Note that you should use ``totalview --args flux mini attach`` when your
job has been launched via ``flux mini run`` or ``flux mini submit``.
In other words, Flux currently does not allow TotalView
```

debugging.rst
```diff
@@ -0,0 +1,84 @@
...
+
+  totalview --args flux mini run -N 2 -n 2 ./mpi-program
+
+Attaching to an already running job:
+
+.. code-block:: sh
+
+  totalview --args flux mini attach <JOBID>
+
+Note that you should use ``totalview --args flux mini attach`` whether your
```
SteVwonder, line 27: ReSt has a note directive, if you want to use that:

```suggestion
.. note::
You should use ``totalview --args flux mini attach`` whether your
```

dongahn:
@petertea: Moving the discussion to GitHub PR. 

> A quick comment.  The links for TotalView still point to the RogueWave site,
which do still work, but I wonder if you might want to update those to reflect
Perforce and the new web site,

> https://totalview.io

Yes I will do this. 

> The doc links still works since we haven't moved those (yet).

> I'm trying a test on butte, where a new version of Spectrum MPI has been
installed, and I'm getting a Unable to connect to Flux, and it's an errno 2,
ENOENT, so it seems like it's not finding o   


> This is just trying to run flux.
I haven't brought TotalView into the mix yet.   I'll take a closer look at the
complete documentation to see if I'm missing something obvious.

You might find the following doc useful: https://flux-framework.readthedocs.io/en/latest/coral.html


> Overall the addition makes sense to me as how to run TotalView. The
exec_handling string might need a bit of work to make it more compact, but I'd
need to play around with it a bit before I can make a recommendation.

Cool.

conf.py
dongahn: Yes, too much churn if we have to update this each time. The version number shows up in the browser window and it was confusing. 

Do we need a way to sync the html doc version with flux-core version though?

dongahn: @SteVwonder and @gonsie: just recommend what you need with the version number, I will adjust the PR.

debugging.rst
dongahn: Learned a new thing! Will change.

debugging.rst
dongahn: OK. Will try that.

debugging.rst
dongahn: Good idea. Just to make sure: here, the external hyperlink is given with the "normal" syntax?

debugging.rst
dongahn: Nice! Pretty powerful. Does ReST support most of GitHub mark down verbatim block ? In this case, for example ```console?

debugging.rst
dongahn: Cool!

conf.py
SteVwonder: > Do we need a way to sync the html doc version with flux-core version though?

From previous conversations in coffee time and issues, I think the conclusion we came to is:
"Each repo has it’s own documentation and that gets tagged and released along with the code, but the high-level “meta” documentation has it’s own versioning that is divorced from any particular sub-packages/repos versioning."

So the top-level docs can refer to either the latest version of flux-core docs or a specific tag (if necessary) via Sphinx links.  But the versioning of flux-docs itself is entirely independent.  We still need to come up with conventions for how and when we want to tag flux-docs.

SteVwonder: > @SteVwonder and @gonsie: just recommend what you need with the version number, I will adjust the PR.

What happens if we just wholesale delete the line of code like was done in flux-core? I think it should just not version anything, and we will just get `latest`.  Which seems fine to me for now until we come up with a convention for tagging and releasing flux-docs.  Thoughts?

dongahn: > So the top-level docs can refer to either the latest version of flux-core docs or a specific tag (if necessary) via Sphinx links. But the versioning of flux-docs itself is entirely independent. We still need to come up with conventions for how and when we want to tag flux-docs.

Very wise decision! Thanks.

debugging.rst
SteVwonder: The Sphinx documentation describes it better than I can: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links

There is basically two ways to do it: with an inline/anonymous link or with a "ref" and "target definition". For this particular case, I recommend trying the latter since this link it used multiple times. You can define the target once, and then reference it multiple times.

dongahn:
@SteVwonder or @gonsie: This PR should be ready. I addressed all of your comments and incorporated other findings that came out from testing and talking to TotalView folks. This is based on some changes in [this PR](https://github.com/flux-framework/flux-core/pull/3130). So it would be wise to land that PR first.

debugging.rst
```diff
@@ -0,0 +1,101 @@
...
+This means you must manually click "Yes" whenever a stop-at-exec
+dialogue window is popped up. You do not need to stop
+it unless you are a Flux developer!
+Instead, you can make TotalView automatically
+continue on these events by building the the following
+exec-handling Tcl code into TotalView.
+
+.. code-block::
+
+    catch {dset TV::exec_handling { {^(flux|lrun|srun|jsrun)(<python[^>]*>(<flux-job>)?|<bash>|<jsrun>|<jswrap>)*$ go}} }
```
SteVwonder, line 69: Unfortunately, this block is not rendering in readthedocs right now.  I wonder if it is because the language is missing (although the docs say in Sphinx 2.0 it became optional):

```suggestion
exec-handling Tcl code into TotalView.

.. code-block:: tcl

    catch {dset TV::exec_handling { {^(flux|lrun|srun|jsrun)(<python[^>]*>(<flux-job>)?|<bash>|<jsrun>|<jswrap>)*$ go}} }
```

If that doesn't work, it looks like you can get a literal block without the code block directive by ending the previous line with `::` and indenting the actual block [[1]](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks):

```suggestion
exec-handling Tcl code into TotalView::

    catch {dset TV::exec_handling { {^(flux|lrun|srun|jsrun)(<python[^>]*>(<flux-job>)?|<bash>|<jsrun>|<jswrap>)*$ go}} }
```

debugging.rst
```diff
@@ -0,0 +1,101 @@
+.. _debugging:
+
+==============
+Debugging Jobs 
+==============
+
+Flux supports parallel debuggers such as Rogue Wave Software (RWS)'s
+TotalView parallel debugger (https://totalview.io).
```
SteVwonder, line 8: Nitpick. Feel free to ignore:

```suggestion
`TotalView parallel debugger <(https://totalview.io)>`_.
```

debugging.rst
dongahn: I tried with Tcl and it didn't work. I will just go with the second suggestion.

dongahn:
OK. Made those suggested changes. I just also put MWP label.

PR #57, opened 2020-08-08
Add Flux batch job HOWTO section

dongahn:
Add doc on how to submit your batch scripts with
some examples.

Use single-user Flux examples with a caveat that
same commands and techniques will also work at the system level.

### DIFF
--- a/batch.rst
+++ b/batch.rst
@@ -1,0 +1,218 @@
+.. _batch:
+
+==========
+Batch Jobs
+==========
+
+In Flux, a batch job is a script submitted along
+with a request for resources. When resources become
+available, the script is run as the
+`initial program <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_8.html#initial-program-program-1>`_ of a new single-user instance of Flux,
+with the batch user as `instance owner <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_8.html#terminology>`_.
+Thus, in contrast to a traditional batch scheduler, a Flux batch job
+has access to fully featured resource manager instance, and may not only
+serially execute work on allocated resources, but load custom modules
+or customize parameters of existing modules, monitor
+and update status of allocated resources, and even submit
+more batch jobs to run further sub-instances of Flux.
+
+.. note::
+   In the following, batch jobs will be described using single-user Flux,
+   but it is important to note that the same commands and techniques
+   will also work at the system level.
+
+----------------------------------
+Launching Flux in Interactive Mode
+----------------------------------
+
+For this demonstration, we first launch Flux under SLURM and get an interactive shell.
+
+.. code-block:: console
+
+  $ salloc -N4 -ppdebug
+  salloc: Granted job allocation 5620626
+  $ srun -N ${SLURM_NNODES} 4 -n ${SLURM_NNODES} --pty --mpi=none --mpibind=off flux start
+
+.. note::
+   If launching under SLURM is not possible or convenient, a single-node
+   single user instance of Flux can be started with ``flux start -s 4``.
+
+------------------
+Mini Batch Command
+------------------
+
+The high-level Flux command that users can use to submit a
+batch script is ``flux mini batch``.
+
+``sleep_batch.sh``:
+
+.. code-block:: sh
+
+  #!/bin/bash
+  
+  echo "Starting my batch job"
+  echo "Print the resources allocated to this batch job"
+  flux hwloc info
+  
+  echo "Use sleep to emulate a parallel program"
+  echo "Run the program at a total of 4 processes each requiring"
+  echo "1 core. These processes are equally spread across 2 nodes."
+  flux mini run -N 2 -n 4 sleep 120
+  flux mini run -N 2 -n 4 sleep 120
+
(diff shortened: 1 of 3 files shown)


batch.rst
```diff
@@ -0,0 +1,156 @@
...
+==========
+Batch Jobs
+==========
+
+Flux supports batch jobs. Users can submit batch scripts
+into Flux, which are queued up and then run when they can be
+allocated to the requested resources. Flux, however, strays away
+from the traditional approach in that batch jobs are
+supported not only at the multi-user system level
+but also at the nested single-user levels.
```
grondo, line 12: I wonder if we should introduce what we mean by a "batch job" in this opening section.

By that I mean something like (perhaps needs more wordsmithing)

> In Flux, a batch job is a script submitted along with a request for resources. When resources become available, the script is run as the *initial program* (alternately: "with access to the resources") of a new single-user instance of Flux, with the batch user as *instance owner*. Thus, in contrast to a traditional batch scheduler, a Flux batch job has access to fully featured Flux instance, and may not only serially execute work on allocated resources, but may load custom modules or customize parameters of existing modules, monitor and update status of allocated resources, and even submit more batch jobs to run further sub-instances of Flux.

batch.rst
```diff
@@ -0,0 +1,156 @@
...
+   In the following, batch jobs will be described using single-user Flux,
+   but it is important to note that the same commands and techniques
+   will also work at the system level.
+
+--------------
+Launching Flux
+--------------
+
+For this demonstration, we first launch Flux under SLURM and get an interactive shell.
+
```
grondo, line 24: Add note:

> Note: If launching under Slurm is not possible or convenient, a single-node single user instance of Flux can be started with `flux start -s 4`.

dongahn:
From the initial feedback from Xiaohua, there are a few things that I'd like to add to this:

1. Explain one needs to drop `--pty` for non-interactive use and add `rundir` option to redirect content store
2. The user was confused that they needed to use either a HERE doc or a separate script to flux start for non-interactive use. I will add this. 
3. Use of `flux queue drain`
4. Blocking vs. non-blocking (e.g., don't background `flux mini batch`; the semantics of flux mini run is foreground although backgrounding it is supported).

batch.rst
dongahn: Great suggestion! I will add this or similar.

batch.rst
dongahn: I wonder adding this to the note may mislead some folks though. Is this something we want to advertise to end users or is this mostly for testing?

grondo: I think it is very useful for playing around with Flux interfaces without "wasting" batch resources or if you do not have easy access to Slurm or other batch launcher. I leave it up to you if you want to include this or not.

However, I will say I believe this single node method to start Flux for testing or experimentation is one of the great features that sets Flux's usability above other RMs. I think normal users could end up using it more than we might initially think.

dongahn: Perhaps we should add a note on the availability of this stand alone mode as well as clear differences between srun `flux start` vs. flux start. 

I remember some users try `flux start`with no starter process within a batch allocation and got confused. And I also remember they added `flux start -s <N>` with the start processes and got confused.

A fuller note should be useful after all. 

Thanks!

jameshcorbett: Yeah, even if it is just for testing, I think it is pretty useful to be able to execute ``flux start flux mini run -n4 hostname`` and similar commands from a login node. I think there are plenty of users who aren't very good at getting interactive batch sessions (plenty don't know about ``salloc`` at all), so it's especially important. Most of the important concepts you can get just by fooling around on a login node without involving anything heavyweight. I suppose there is some danger that someone would try to run an HPC application on the login node, though...

And yeah, I think one of the confusing things for users will be the difference between ``flux start CMD`` and ``flux start``, because the second one makes it _look_ like nothing happened---like flux is suddenly just running in the same shell as before---and if you just copy your terminal session into a batch script, it won't work as expected.

dongahn: Thanks @jamescorbett for chiming in!

jameshcorbett:
This looks great, Dong. My only comments are things you've already pointed out:

I think it would be good to point out that there are two ways, blocking and nonblocking, to submit batch jobs and regular jobs: alloc, batch, run, and submit. 

You might also want to provide a full example of using Flux on Slurm, which I imagine might look something like:

outer_script.sh:
```
#!/bin/sh
#SBATCH -N 8
#SBATCH etc

srun ... flux start inner_script.sh
```

inner_script.sh:
```
#!/bin/sh
flux mini submit -n5 hostname
flux mini submit -n17 spam --foo --bar

flux queue drain
```

And you submit this to Slurm with `sbatch outer_script.sh`. When you already have a Flux instance going, this would be much simpler---get rid of `outer_script.sh` and just execute `flux mini batch -n8 -N8 -c? inner_script.sh`. In fact, I think it may be worth going over that difference.

dongahn:
OK. I have addressed all of the reviewers' (great) comments and incorporated the finings from working with users. I squashed all the interim commits. From my perspective, this is good to go in.

dongahn:
Ok. I resolved the conflicts and forced a push. It will be good if we can merge this sooner rather than later. We already have some users needing this doc.

PR #62, opened 2020-09-16
Update CORAL-specific instructions for latest Spectrum updates

SteVwonder:
The latest Spectrum version requires a flag to disable the loading of a pami gpu library that segfaults when not run directly under jsrun (e.g., when Flux is bootstrapping).

Also include the instruction to load our `flux` module to access side installs of Flux.

I manually tested the updated instructions on both Lassen and Summit.  I don't think the `-disable_gpu_hooks` flag does anything on Summit since they are running an older Spectrum version, but thankfully, `jsrun` and Spectrum MPI do not complain when it is provided.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,11 +14,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - uses: ammaraskar/sphinx-action@master
-      with:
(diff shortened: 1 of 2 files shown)


cmoussa1 (approved):
LGTM! Thanks @SteVwonder!

grondo:
Github workflow config was changed so this needs a manual merge. I'll take care of that.

PR #65, opened 2020-09-16
readthedocs: add top-level ReadTheDocs (RTD) config yaml file

### LINKED ISSUE #64: Admin Guide: syntax highlighting in TOML examples
It looks like the admin guide uses TOML code blocks in the ReST code, and the syntax highlighting of those blocks works on GitHub but not on ReadTheDocs.  Not sure if this reproduces locally, but it happens on the hosted RTD site.

GitHub Screenshot:
Screen Shot 2020-09-16 at 12 38 44 PM

RTD Screenshot:
Screen Shot 2020-09-16 at 12 38 52 PM

SteVwonder:
Problem: RTD switched from primarily configuring builds via their web interface to using version control yaml files. Additionally, the build is currently using an old version of Pygments that does not support syntax highlighting of toml.

Solution: kill two birds with one stone by adding an RTD top-level config yaml that requests python 3.7 and then request a newer version of Pygments (which requires py3.5+).

Closes #64

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,16 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the top-level directory with Sphinx
+sphinx:
+  configuration: conf.py
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.7
+  install:
+    - requirements: requirements.txt
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+Pygments>=2.4
 Sphinx
 sphinxcontrib-spelling
 sphinx-rtd-theme


grondo (approved):
Looks great to me. Awesome, thanks!

SteVwonder:
Doesn't seem like a newer Pygments was pulled in just by asking for py3.7.  It looks like RTD pins Sphinx and Pygments to specific versions.  I wonder if we can override those.

grondo:
Hm, looking at the rendered PR from the `docs/readthedocs.org` check above the TOML is still not highlighted. I'm going to assume that is because the config added here will only be used after it is merged into the main branch.

grondo:
Ah, I noticed that but thought perhaps the config wasn't pulled in for RTD PR builds. (Of course I was lazy and didn't check the build logs.)

SteVwonder:
There we go! Forcing `Pygments>=2.4` (which is the version that added TOML support), fixes the syntax highlighting issue.  Hopefully there are no objections to adding that line to the `requirements.txt`.  If there are, I can always make a separate `requirements-rtd.txt` just for RTD that has that constraint.

SteVwonder:
I think I'm happy with this PR now.  Since I made changes after you approved, are you still OK with this PR @grondo?

grondo:
Yeah, awesome thanks!

PR #66, opened 2020-09-19
[WIP] coral: update GPU section to include new `flux hwloc info` GPU info

SteVwonder:
Problem: it can be difficult to tell if Flux has detected GPUs on the
nodes that it is running on.

Solution: suggest that users run `flux hwloc info` to see if Flux
detected any GPUs

Marking this as a WIP since `flux resource list` should really be used rather than `flux hwloc info`.


SteVwonder:
This has been replaced by #87

PR #74, opened 2020-11-04
adminguide: explain how to configure resources

garlick:
Rename the "Resource exclusion" section to "Resource configuration",
and add an explanation of the 'path' key and an example command line
for generating R with flux-R.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -244,26 +244,39 @@ node along with other critical cluster services.
 
 .. _configuration-resource-exclusion:
 
-^^^^^^^^^^^^^^^^^^
-Resource exclusion
-^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^
+Resource configuration
+^^^^^^^^^^^^^^^^^^^^^^
 
-It may be desirable to prevent resources on management and login nodes from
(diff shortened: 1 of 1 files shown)


adminguide.rst
```diff
@@ -244,27 +244,40 @@ node along with other critical cluster services.
...
+Resource configuration
+^^^^^^^^^^^^^^^^^^^^^^
+
+The system resource configuration may be generated in RFC 20 (R version 1)
+form using ``flux R encode``.  At minimum, a hostlist and core idset must
+be specified on the command line, e.g.
+
+.. code-block:: console
+
+ $ flux R encode -H fluke[3,108,6-103] -c0-3 > /etc/flux/system/R
```
grondo, line 257: It might be more illustrative to use long options in the example 
```suggestion
 $ flux R encode --hosts=fluke[3,108,6-103] --cores=0-3 > /etc/flux/system/R
```

grondo (approved):
LGTM. Just one suggestion inline that is take-it or leave-it.

adminguide.rst
garlick: Good idea!

grondo:
Looks good, I'll set MWP.

PR #80, opened 2020-12-16
adminguide: update for flux-core release 0.22.0

garlick:
Bring most of the 0.21.0 warnings forward to 0.22.0.

Change the flux-keygen usage and CURVE cert path.

Add curve_cert to [bootstrap] configuration.

Change exclusions to use a hostlist.

Change drain to use a hostlist and use 'flux resource drain' with
no arguments to list drained nodes instead of 'flux resource status'.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -13,12 +13,12 @@ resource manager on a cluster.
     in this guide may change with regularity.
 
     This document is in DRAFT form and currently applies to flux-core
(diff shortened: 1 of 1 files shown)


grondo (approved):
Looks good to me! We should probably mention `flux resource status` as the way to check on excluded, drained, and offline nodes/targets in one command. I could look at adding this as a follow on PR though.

PR #81, opened 2020-12-16
adminguide: add description of flux-resource status and list commands

grondo:
Add a section introducing the flux-resource `status` and `list` commands, so that admins are at least aware of these tools and their potential use.


garlick (approved):
Great addition!  Thanks for taking the time to write that.

grondo:
Thanks!

PR #93, opened 2021-03-24
adminguide: fix typo in toml path

SteVwonder:
Replace errant `/` with `.` for file extension name


grondo (approved):
Good catch, thanks!

PR #95, opened 2021-04-06
faqs: change openmpi version number 4.0.2 -> 4.0.1

garlick:
Problem: faq states that 4.0.2+ works, but also says a bug
affected versions _through_ 4.0.2.

I think the bug affects 4.0.1 but not 4.0.2, so state that the
bug affects versions through 4.0.1.


SteVwonder:
Thanks for catching that @garlick!  Sorry for the bug.

SteVwonder (approved):

PR #98, opened 2021-08-17
Add stats quick start page

garrettbslone:


garrettbslone:
Added `django` to dictionary and force pushed for CI.

stats.rst
```diff
@@ -0,0 +1,209 @@
...
+
+  $ export FLUX_FRIPP_STATSD=hostname:port
+
+The second thing needed is to run brubeck. There are a few option, but no real best way as of now.
+
+.. code-block:: console
+
+  # launch brubeck as a background job in flux
+  $ flux start
+  $ flux mini run ~/path/to/brubeck --config=~/path/to/config.default.json
```
grondo, line 152: Do you mean `flux mini submit`? `flux mini run` will block until brubeck exits.

stats.rst
```diff
@@ -0,0 +1,209 @@
+.. _stats:
+
+=================
+Stats Quick Start
+=================
+
+A quick guide to getting the internal stats collection working in flux.
+This guide assumes you have already built a flux binary; if you have not see :ref:`quickstart`.
```
grondo, line 8: A larger summary here of what is meant by "stats" (e.g. counters, gauges, and timers) and a reference to StatsD might be useful to give a new user some context here.

stats.rst
```diff
@@ -0,0 +1,209 @@
...
+  $ cp config.default.json.example config.default.json
+  $ make test
+
+If all of the tests pass then it's time to move on to the frontend and third piece of the stack, Graphite.
+
+-------------------
+Installing Graphite
+-------------------
+
+Install Graphite from your local package manager.
```
grondo, line 33: Should it also be mentioned that you can run graphite in docker? Or does that not work for some reason?

stats.rst
```diff
@@ -0,0 +1,209 @@
...
+
+First make sure that the carbon cache is enable on boot.
+
+.. code-block:: console
+
+  $ sudo vim /etc/default/graphite-carbon
+  # set to true
+  CARBON_CACHE_ENABLED=true
+
+Next, you will need to edit carbon's config file and set the specific you want. You can fin specifics on the configurations `here <https://graphite.readthedocs.io/en/stable/config-carbon.html>`_.
```
grondo, line 182: Some typos in this line "and set the specific you want" and "You can fin specifics on the configurations"

grondo (commented):
Just a few minor things found on a quick pass. Thanks!

stats.rst
garrettbslone: All of the docker images I've found with Graphite also contain statsd which would work fine unless someone is using the `flux_stats_count()`

grondo: You could just run brubeck on a different port or not export the statsd udp port on the docker line. However, I can experiment with that and update the docs down the road.

garrettbslone: For now should I add a note about docker at the top?

grondo: You could mention it -- however making the config changes you mention below in the docker image may be tricky.

Later on, maybe we'll make our own `fluxrm/graphite` docker image with your suggested configuration and a copy of brubeck instead of statsd -- but don't worry about it too much for now.

garrettbslone:
Force pushed to make changes requested above.

stats.rst
grondo: Interesting, I just learned that the latest `graphiteapp/graphite-statsd` docker image has brubeck installed: 

https://github.com/graphite-project/docker-graphite-statsd#brubeck

garrettbslone:
Added docker setup to stats page and force pushed.

stats.rst
```diff
@@ -0,0 +1,308 @@
+.. _stats:
+
+=================
+Stats Quick Start
+=================
+
+A quick guide to getting the internal stats collection working in flux.
+This guide assumes you have already built a flux binary; if you have not, see :ref:`quickstart` to get started.
```
grondo, line 8: You can probably just delete these two statements (the first isn't a sentence).

I'd suggest replacing with a sentence that describes what "stats" is, e.g.

> Flux includes an optional instrumentation facility to collect internal metrics, or "stats", and send them to an aggregator for eventual display in a backend graphing tool such as Graphite.

stats.rst
```diff
@@ -0,0 +1,308 @@
...
+The internal stats collection in flux is built on the `statsd <https://github.com/statsd/statsd>`_ protocol and
+following the typing convention used by `brubeck <https://github.com/lukepalmer/brubeck>`_.
+Flux stats supports three types: counters, gauges, and timings.
```
grondo, line 11:
```suggestion
Flux stats supports three metrics types: counters, gauges, and timings.
```

stats.rst
```diff
@@ -0,0 +1,308 @@
...
+ * Gauge   - An integer that takes an arbitrary value and maintains its value until it is set again. Gauges can also take incremental values in the form of a + or - in front of the value which increments the previously stored value. An example of where to use a gauge is to track the current size of the broker's content-cache. At each point, the cache's sizes are independent of each other.
+ * Timing  - A double value which represents the time taken for a given task in ms. An example of where to use a timer is timing the length of asynchronous loads in the broker's content-cache. The cache entry can keep track of when the load was started and then calculate and send the time taken once the entry is loaded.
+
+For more information on statsd types see the `statsd docs <https://github.com/statsd/statsd/blob/master/docs/metric_types.md>`_.
+
+---------------------------------
+Running Brubeck/Graphite Manually
+---------------------------------
```
grondo, line 21: I'd switch the order to have the Docker instructions first.

stats.rst
```diff
@@ -0,0 +1,308 @@
...
+  # catch all metrics that start with 'flux'
+  # and retain each data point as 1s
+  [flux]
+  pattern = ^flux.*
+  retentions = 1s:3d
+
+  # remove the old whisper files
+  $ cd $(docker inspect graphite | \
+    jq '.[].Mounts[] | select(.Destination == "/opt/graphite/storage")' | \
+    awk '/Source/ {print substr($2, 2, length($2) - 3)}') && \
```
grondo, line 248:
```suggestion
  $ cd $(docker inspect graphite | \
    jq -r '.[].Mounts[] | select(.Destination == "/opt/graphite/storage") | .Source' |  && \
```

grondo (commented):
I have to run an errand, but added some more quick comments.

stats.rst
```diff
@@ -0,0 +1,307 @@
+.. _stats:
+
+=================
+Stats Quick Start
+=================
+
+Flux includes an optional instrumentation facility to collect internal metrics, or "stats",
+and send them to an aggregator for eventual display in a front end graphing tool such as Graphite.
+The internal stats collection in flux is built on the `statsd <https://github.com/statsd/statsd>`_ protocol and
+following the typing convention used by `brubeck <https://github.com/lukepalmer/brubeck>`_.
```
grondo, line 10:
```suggestion
follows the typing convention used by `brubeck <https://github.com/lukepalmer/brubeck>`_.
```

stats.rst
```diff
@@ -0,0 +1,307 @@
+.. _stats:
+
+=================
+Stats Quick Start
+=================
+
+Flux includes an optional instrumentation facility to collect internal metrics, or "stats",
+and send them to an aggregator for eventual display in a front end graphing tool such as Graphite.
+The internal stats collection in flux is built on the `statsd <https://github.com/statsd/statsd>`_ protocol and
```
grondo, line 9: I think we've settled on treating Flux the project as a proper noun, i.e. capitalized, vs `flux` the command with no capitalization, i.e.:

```suggestion
The internal stats collection in Flux is built on the `statsd <https://github.com/statsd/statsd>`_ protocol and
```

stats.rst
```diff
@@ -4,14 +4,89 @@
 Stats Quick Start
 =================
 
-A quick guide to getting the internal stats collection working in flux.
-This guide assumes you have already built a flux binary; if you have not see :ref:`quickstart`.
+Flux includes an optional instrumentation facility to collect internal metrics, or "stats",
+and send them to an aggregator for eventual display in a front end graphing tool such as Graphite.
+The internal stats collection in flux is built on the `statsd <https://github.com/statsd/statsd>`_ protocol and
+following the typing convention used by `brubeck <https://github.com/lukepalmer/brubeck>`_.
+Flux stats supports three metric types: counters, gauges, and timings.
```
garlick, line 11: s/following/follows/

And suggest you for the metric types, you drop the plural, e.g. counter, gauge, and timing.

stats.rst
```diff
@@ -4,14 +4,89 @@
...
-------------------
+For more information on statsd types see the `statsd docs <https://github.com/statsd/statsd/blob/master/docs/metric_types.md>`_.
+
+------------
+Using Docker
+------------
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+Running Brubeck & Graphite
+^^^^^^^^^^^^^^^^^^^^^^^^^^
```
garlick, line 25: Seems like the major heading here should be "running brubeck and graphite" and the minor ones "using docker" and, later, "using git/package manager" or similar.

Also, at the beginning of the "running brubeck and graphite" section, a short intro might be useful, like:
> Two methods for starting brubeck for stats aggregation and graphite for stats display are presented below.   The docker method is recommended, if docker is available.  Both methods configure brubeck to accept stats messages from Flux on UDP port 8126, and graphite to capture stats from brubeck and present them via a web server  at  `http://127.0.0.1`.

stats.rst
```diff
@@ -0,0 +1,307 @@
...
+Stats Quick Start
+=================
+
+Flux includes an optional instrumentation facility to collect internal metrics, or "stats",
+and send them to an aggregator for eventual display in a front end graphing tool such as Graphite.
+The internal stats collection in flux is built on the `statsd <https://github.com/statsd/statsd>`_ protocol and
+following the typing convention used by `brubeck <https://github.com/lukepalmer/brubeck>`_.
+Flux stats supports three metric types: counters, gauges, and timings.
+
+ * Counter - An integer value that will, on the back end (brubeck) send the count and reset to 0 at each flush. It calculates the change from the value sent at the previous flush. An example of where to use a counter is the builtin message counters that are part of each flux_t handle. The counts will continually increase and brubeck will handle calculating the count sent in the interval.
```
grondo, line 13:
```suggestion
 * Counter - An integer value that will, on the back end (brubeck), send the count and reset to 0 at each flush. It calculates the change from the value sent at the previous flush. An example of where to use a counter is the builtin message counters that are part of each ``flux_t`` handle. The counts will continually increase and brubeck will handle calculating the count sent in the interval.
```

stats.rst
```diff
@@ -0,0 +1,307 @@
...
+ * Gauge   - An integer that takes an arbitrary value and maintains its value until it is set again. Gauges can also take incremental values in the form of a + or - in front of the value which increments the previously stored value. An example of where to use a gauge is to track the current size of the broker's content-cache. At each point, the cache's sizes are independent of each other.
+ * Timing  - A double value which represents the time taken for a given task in ms. An example of where to use a timer is timing the length of asynchronous loads in the broker's content-cache. The cache entry can keep track of when the load was started and then calculate and send the time taken once the entry is loaded.
+
+For more information on statsd types see the `statsd docs <https://github.com/statsd/statsd/blob/master/docs/metric_types.md>`_.
+
+------------
+Using Docker
+------------
```
grondo, line 21: Suggestion: have a section for "Running brubeck and Graphite"
with subsections for "Using Docker" and "Installing from packages and source" (Or something)

stats.rst
```diff
@@ -0,0 +1,307 @@
...
+
+.. code-block:: console
+
+  $ sudo systemctl restart carbon-cache
+
+--------------------------------
+Running Stats Collection in Flux
+--------------------------------
+
+Once you have brubeck and Graphite setup and Graphite running there are only two more things left to be able to get stats from ``flux``.
```
grondo, line 257:
```suggestion
Once you have brubeck and Graphite setup and Graphite running there are only two more things left to be able to get stats from Flux.
```

stats.rst
```diff
@@ -0,0 +1,307 @@
...
+First you will need to set the ``FLUX_FRIPP_STATSD`` environment variable with the endpoint of the aggregator (brubeck in this case).
+If you just used the default config for brubeck as suggested earlier, the endpoint is ``0.0.0.0:8126``. If you used the docker setup the
+endpoint is ``0.0.0.0:8125``.
+
+.. code-block:: console
+
+  $ export FLUX_FRIPP_STATSD=hostname:port
```
grondo, line 264: I actually found that `hostname:port` doesn't work, should this be `ipaddr:port`?

e.g.:

```console
$ FLUX_FRIPP_STATSD=localhost:8125 src/cmd/flux start  flux mini submit --watch --wait --progress --cc=0-10 hostname
2021-08-19T15:19:07.329074Z broker.err[0]: error creating server address: No such file or directory
2021-08-19T15:19:07.329129Z broker.err[0]: error creating server address: No such file or directory
2021-08-19T15:19:07.329185Z broker.err[0]: error creating server address: No such file or directory
2021-08-19T15:19:07.329208Z broker.err[0]: error creating server address: No such file or directory
2021-08-19T15:19:07.329230Z broker.err[0]: error creating server address: No such file or directory
2021-08-19T15:19:07.329251Z broker.err[0]: error creating server address: No such file or directory

```

stats.rst
```diff
@@ -0,0 +1,307 @@
...
+
+^^^^^^^^^^^^
+Manual Setup
+^^^^^^^^^^^^
+
+The second thing needed is to run brubeck. There are a few option, but no real best way as of now.
```
grondo, line 270: It should be mentioned here that, when using the graphiteapp docker image, `brubeck` is already running inside of that image. Perhaps this section should be named `Optional: launch brubeck inside of a Flux instance` or something.

I would say you could even replace this section with a simple "Note" that says, "if brubeck is not already running, for example in the graphite docker image or started outside of any Flux instance, then it will need to be started now".

We can perhaps leave the open question of how to start a per-instance brubeck to the reader (especially in the docker case, that will be a little tricky and perhaps beyond the scope of a quickstart guide)

grondo (commented):
A few more comments. Nice job!

stats.rst
```diff
@@ -4,14 +4,89 @@
...
+    -p 2023-2024:2023-2024 \
+    -p 8125:8125/udp \
+    -p 8126:8126 \
+    -e BRUBECK=1 \
+    graphiteapp/graphite-statsd
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Configuring Graphite (Optional)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
garlick, line 44: Perhaps add some explanation here, like
> Graphite's default configuration limits stats recording and display to one update per minute.  If higher resolution is desired, graphite must be reconfigured.

stats.rst
```diff
@@ -207,3 +249,59 @@ Now restart the carbon daemon and the new configuration should take effect.
...
+    --name graphite \
+    --restart=always \
+    -p 80:80 \
+    -p 2003-2004:2003-2004 \
+    -p 2023-2024:2023-2024 \
+    -p 8125:8125/udp \
+    -p 8126:8126 \
+    -e BRUBECK=1 \
+    graphiteapp/graphite-statsd
+
```
garlick, line 307: Seems to be a repeat of earlier section?

garlick (changes requested):
A few comments inline.

Also: looks like the most recent changes got added to the wrong commit.

garlick:
@garrettbslone - apologies, my comments were made before I saw that @grondo had also commented.   We said some of the same things.  Maybe address his comments first and then look over mine to see if anything useful remains.

garrettbslone:
Force pushed to address recent comments.

stats.rst
```diff
@@ -0,0 +1,271 @@
...
+--------------------------
+Running Brubeck & Graphite
+--------------------------
+
+Two methods for starting brubeck for stats aggregation and Graphite for stats display are presented below.
+If possible, it is recommended to use the docker method. Both methods configure brubeck to receive stats
+from Flux on a udp port (Docker uses 8125 and the packages/source setup uses 8126), and
+graphite to receive stats sent from brubeck and display them on a local web server which can be viewed
+at ``htt://127.0.0.1``.
```
garlick, line 27: Missing "p".

stats.rst
```diff
@@ -0,0 +1,271 @@
...
+  $ sudo systemctl restart carbon-cache
+
+--------------------------------
+Running Stats Collection in Flux
+--------------------------------
+
+Once you have brubeck and Graphite setup and Graphite running there are only two more things left to be able to get stats from Flux.
+First you will need to set the ``FLUX_FRIPP_STATSD`` environment variable with the endpoint of the aggregator (brubeck in this case).
+If you just used the default config for brubeck as suggested earlier, the endpoint is ``0.0.0.0:8126``. If you used the docker setup the
+endpoint is ``0.0.0.0:8125``.
```
garlick, line 264: Using `0.0.0.0` (INADDR_ANY) is a bit weird as a `sendto()` address.  Usually one uses this to bind a service to "all interfaces".  I'm not really sure what it means in this context, but it does seem to work!

One vote for `127.0.0.1` here, as "less wierd".  It's possible I'm ignorant about how that address works though!

garlick (commented):
Couple comments inline.

There's also still some bits in the dictionary commit that shouldn't be there.

stats.rst
garrettbslone: from the default brubeck config:

```bash
"samplers" : [
    {
      "type" : "statsd",
      "address" : "0.0.0.0",
      "port" : 8126,
      "workers" : 4,
      "multisock" : true,
      "multimsg" : 8
    }
  ]
```

Not sure if it makes a difference  to use `127.0.0.1`, but I was just using the config as a reference.

grondo: `0.0.0.0` is the INADDR_ANY address, maybe that config is indicating brubeck will _listen_ on all local addresses?

AFAICT, using `0.0.0.0` as a synonym for localhost (`127.0.0.1`) (probably) only works on Linux. This doesn't  comply with any standard, since connecting or sending a packet to `INADDR_ANY` doesn't make any sense in the IP protocol.

I would have never thought to try `0.0.0.0` and I couldn't find where or why or how Linux translates this to `localhost`.

It does mean that a literal `0` can be used as a shortcut for `localhost` though!

```
$ ping 0
PING 0 (127.0.0.1) 56(84) bytes of data.
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.084 ms
64 bytes from 127.0.0.1: icmp_seq=2 ttl=64 time=0.050 ms
^C
```

Just thinking of the time I've wasted typing l-o-c-a-l-h-o-s-t in my life gives me shivers.

grondo: BTW, I would definitely use `127.0.0.1` for localhost in these examples. Use of `0.0.0.0` is sure to confuse somebody (well it did!) :-)

garlick: By the way @garrettbslone - I should have mentioned this earlier, but a guy I went to school with write an excellent tutorial on socket programming:

https://beej.us/guide/bgnet/html/

Also author of the internet pizza server (try the garlick topping!)

https://beej.us/pizza/

garlick (approved):
I think this is in good shape now.  Are you feeling ready on this one @garrettbslone ?

garrettbslone:
> I think this is in good shape now. Are you feeling ready on this one @garrettbslone ?

Yep, I'm ready to merge it!

PR #109, opened 2021-10-20
admin guide: add flux-accounting setup section

cmoussa1:
This PR adds a new section to the bottom of the Admin Guide that details setup instructions for the flux-accounting database and multi-factor priority plugin as well as configuration of the automatic update scripts for fairshare and job usage values.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -53,7 +53,9 @@ Package installation
 A Flux system install requires ``flux-security`` and ``flux-core``
 packages to be installed at a minimum. For real workloads it is highly
 recommended that ``flux-sched`` (a.k.a the *Fluxion* graph-based scheduler)
-be installed as well.
+be installed as well. For managing user accounts, banks, and utilizing job
+priority calculation and fairshare, the ``flux-accounting`` package should also
+be installed.
 
 If installing from your Linux distribution package manager (preferred),
 e.g. RPM or dpkg, you may skip this section.
@@ -319,6 +321,124 @@ job queue and record of past jobs.
     0.30.0 limitation: Running out of space is not handled gracefully.
     If this happens it is best to stop Flux, remove the content.sqlite file,
     and restart.
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+flux-accounting configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If ``flux-accounting`` was installed on the management node, its database should
+be initialized (its default location is ``/var/lib/flux/FluxAccounting.db``),
+and the job manager configured to automatically load the multi-factor priority
+plugin provided by flux-accounting. All commands presented in the following
+subsections should be run as the flux user.
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Database Creation and Initialization
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The first thing to configure when first setting up the flux-accounting database
+is to set the **PriorityUsageResetPeriod** and **PriorityDecayHalfLife**
+parameters. Both of these parameters represent a number of weeks by which to
+hold usage factors up to the time period where jobs no longer play a factor in
+calculating a usage factor. If these parameters are not passed when creating the
+DB, **PriorityDecayHalfLife** is set to 1 week and **PriorityUsageResetPeriod**
+is set to 4 weeks, i.e the flux-accounting database will store up to a month's
+worth of jobs broken up into one week chunks:
+
+.. code-block:: console
+
+ $ flux account create-db --priority-decay-half-life=2 --priority-usage-reset-period=8
+
+Now that the database is created, as the admin who has permission to write to
+the database, we can initialize it with some banks by specifying a name and the
+number of shares:
+
+.. code-block:: console
+
+ $ flux account add-bank root 1
+ $ flux account add-bank --parent-bank=root sub_bank_A 1
+
+We can then add some users to those banks so those users can run jobs on the
+system:
+
+.. code-block:: console
+
+ $ flux account add-user --username=user1234 --bank=sub_bank_A
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Loading the multi-factor priority plugin
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The other component to load is the multi-factor priority plugin, which can be
+loaded by specifying it in the ``job-manager.plugins`` section of the Flux TOML
+configuration file. Below is an example configuration file:
+
+``/etc/flux/system/conf.d/plugins.toml``
+
+.. code-block:: toml
+
+ [job-manager]
+ plugins = [
+   { load = "/usr/lib64/flux/job-manager/plugins/mf_priority.so" },
+ ]
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Setting up the automatic update scripts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are three update commands and scripts that require setup in order for the
+database to be periodically updated with new job usage and fairshare
+information. These commands and scripts should be run as the flux user and they
+should be run in order and with the same frequency to ensure accurate updates.
+A suggested frequency for running these automatic updates would be to run these
+once every 30 minutes, but can be tuned to run more or less frequently.
+
+The first automatic cron job that should run is the ``update-usage`` subcommand.
+This subcommand fetches the most recent job records from the job-archive
+database for every user in the flux-accounting database and calculates a new job
+usage value. This subcommand takes one optional argument,
+``--priority-decay-half-life`` which should be set to the same value as the
+``--priority-decay-half-life`` parameter mentioned above when first creating the
+database; it represents the number of weeks to hold one job usage "chunk."
+If not specified, this optional argument also defaults to 1 week.
+
+.. code-block:: console
+
+ $ flux account update-usage
+
+After the job usage values are re-calculated and updated, the fairshare values
+for each user also need to be updated. This can be accomplished by configuring
+the ``flux-update-fshare`` command to also run as the second cron job. This
+fetches user account data from the flux-accounting DB and recalculates and
+writes the updated fairshare values back to the DB.
+
+.. code-block:: console
+
+ $ flux account-update-fshare
+
+Once the fairshare values for all of the users in the flux-accounting DB get
+updated, this information will be sent to the priority plugin. The
+``flux account-priority-update`` command can also be configured to run as the
+third cron job to send the updated values to the plugin:
+
+.. code-block:: console
+
+ $ flux account-priority-update
+
+Together, the three automatic update scripts can be combined to run as the flux
+user in succession of one another via ``cron`` using a shell script like the
+following example:
+
+``flux-accounting-update.sh``
+
+.. code-block:: sh
+
+ #!/bin/bash
+
+ flux account update-usage
+ flux account-update-fshare
+ flux account-priority-update
+
 
 ------------------------------
 System Instance Administration
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -489,3 +489,5 @@ El
 Capitan
 Perlmutter
 glibc
+cron
+fairshare


adminguide.rst
```diff
@@ -771,3 +771,44 @@ circular buffer.  For some problems, it may be useful to view this log:
...
+
+.. code-block:: console
+
+ $ flux account create-db --priority-decay-half-life=2 --priority-usage-reset-period=8 path/to/DB
+
+The other component to load is the multi-factor priority plugin, which can be loaded with ``flux jobtap load``:
+
+.. code-block:: console
+
+ $ flux jobtap load mf_priority.so
```
grondo, line 792: Suggestion, also describe how to load the `mf_priority.so` via configuration, which will be how administrators will want to configure the cluster to avoid running manual commands on start/restart.

adminguide.rst
```diff
@@ -771,3 +771,44 @@ circular buffer.  For some problems, it may be useful to view this log:
...
+
+.. code-block:: console
+
+ $ flux update-fshare -f path/to/DB
+
+Once the fairshare values for all of the users in the flux-accounting DB get updated, this information will be sent to the priority plugin. The ``bulk_update.py`` script can also be configured to run as a cron job to send the updated values to the plugin:
+
+.. code-block:: console
+
+ $ flux python bulk_update.py -p path/to/DB
```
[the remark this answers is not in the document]
cmoussa1: Ah, so I did some digging yesterday afternoon, and it seems I neglected to add this python script to `dist_fluxcmd_SCRIPTS` in the Makefile.am. Thus, when an RPM is built, this script doesn't get installed and won't be able to be called with `flux`. :/ That's my mistake. 

I went ahead and made this correction yesterday by moving it to the `src/cmd/` directory and adding it to the list of scripts under `dist_fluxcmd_SCRIPTS`. I also renamed it to more closely resemble its role with the priority plugin, `flux-mf-priority-update.py`. I'll post a PR in flux-accounting soon to address this fix. I guess this fix won't make its way into the built RPM until flux-accounting `v0.11.0` is released at the end of this month. I'm sorry about that.

cmoussa1:
OK, now that flux-framework/flux-accounting#162 has landed, I've revised the flux-accounting configuration section to include the new command names. I also moved the section under the `System Instance Configuration`, as well as included a TOML code block to show how to load the multi-factor priority plugin using a config file (for that, I copied [this example](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html?highlight=plugin); let me know if I got this right).

adminguide.rst
```diff
@@ -770,4 +816,3 @@ circular buffer.  For some problems, it may be useful to view this log:
 ~~~~~~~~~~~~~
 KVS Eventlogs
 ~~~~~~~~~~~~~
-
```
grondo, line 773: Looks like a stray whitespace change here (didn't track down which commit it came from though, sorry)

cmoussa1: 🤦‍♂️ I'm sorry about that; I must have left that over when I moved the accounting section. Just pushed a fix to remove that whitespace change, so that should be fixed now.

garlick:
@cmoussa1, were you able to walk through this in the fluxorama docker image to double check this procedure before we ask our admins to use it on fluke?

Also, was thinking you might want to mention `flux-accounting` in the first paragraph of [package installation](https://flux-framework.readthedocs.io/en/latest/adminguide.html#package-installation) where we introduce `flux-core`, `flux-sched`, and `flux-security`.

cmoussa1:
@garlick I tested this procedure in my interactive `centos8` docker image, which pulls from `fluxrm/flux-core:centos8`. Is this significantly different from the `fluxorama` docker image? If so then I should definitely double check in the `fluxorama` image, thanks for pointing that out.

> Also, was thinking you might want to mention `flux-accounting` in the first paragraph of [package installation](https://flux-framework.readthedocs.io/en/latest/adminguide.html#package-installation) where we introduce `flux-core`, `flux-sched`, and `flux-security`.

Oh, thanks for this suggestion, I can push a change that adds this mention in the first paragraph.

garlick:
> Is this significantly different from the fluxorama docker image?

No I was just thinking it may be easier to try the [job-manager] config in that environment since there is already a config in place.  If you've verified that the steps work then I'm happy!

adminguide.rst
```diff
@@ -320,6 +322,52 @@ job queue and record of past jobs.
     If this happens it is best to stop Flux, remove the content.sqlite file,
     and restart.
 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+flux-accounting configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In order to support fairshare and multi-factor priority ordering of user jobs, the flux-accounting package must be present and configured for the system instance. This allows for the management of users and banks on a cluster, job priorities to be calculated using multiple factors and uses job record information from users to calculate and maintain fairshare values.
```
garlick, line 329: Suggestion: now that flux-accounting was introduced above, perhaps just replace this with
> If `flux-accounting` was installed, its database should be initialized, and the job manager configured to automatically load the multifactor priority plugin provided by `flux-accounting`.

adminguide.rst
```diff
@@ -320,6 +322,52 @@ job queue and record of past jobs.
...
+
+The first thing to configure when first setting up the flux-accounting database is to set the PriorityUsageResetPeriod and PriorityDecayHalfLife parameters. Both of these parameters represent a number of weeks by which to hold usage factors up to the time period where jobs no longer play a factor in calculating a usage factor. If these parameters are not passed when creating the DB, PriorityDecayHalfLife is set to 1 week and PriorityUsageResetPeriod is set to 4 weeks, i.e the flux-accounting database will store up to a month's worth of jobs broken up into one week chunks:
```
garlick, line 331: Is there no other database initialization required?  Add users? Associate banks to users?

Just a thought but maybe we could add subsections here for 1) configure the database for the first time,  2) create TOML config file, and 3) set up cron jobs

adminguide.rst
```diff
@@ -320,6 +322,52 @@ job queue and record of past jobs.
...
+
+.. code-block:: console
+
+ $ flux account create-db --priority-decay-half-life=2 --priority-usage-reset-period=8 path/to/DB
```
garlick, line 335: What node should this run on?  The node designated as the flux rank 0 node?   Are there storage requirements that need to be mentioned?

adminguide.rst
```diff
@@ -320,6 +322,52 @@ job queue and record of past jobs.
...
+
+The other component to load is the multi-factor priority plugin, which can be loaded with ``flux jobtap load``:
+
+.. code-block:: console
+
+ $ flux jobtap load mf_priority.so
+
+This plugin can also be loaded by specifying it in the ``job-manager.plugins`` section of the Flux TOML configuration file. Example:
+
+.. code-block:: toml
```
garlick, line 345: suggest you give this TOML fragment a file name like the other examples in the doc.

adminguide.rst
```diff
@@ -320,6 +322,52 @@ job queue and record of past jobs.
...
+
+ $ flux account create-db --priority-decay-half-life=2 --priority-usage-reset-period=8 path/to/DB
+
+The other component to load is the multi-factor priority plugin, which can be loaded with ``flux jobtap load``:
+
+.. code-block:: console
+
+ $ flux jobtap load mf_priority.so
+
+This plugin can also be loaded by specifying it in the ``job-manager.plugins`` section of the Flux TOML configuration file. Example:
```
garlick, line 343: This is how all system instance configurations will want configure the priority plugin, so it might be appropriate to drop the sentence above that shows how to load it manually.

adminguide.rst
```diff
@@ -320,6 +322,52 @@ job queue and record of past jobs.
...
+
+.. code-block:: console
+
+ $ flux account-update-fshare -f path/to/DB
+
+Once the fairshare values for all of the users in the flux-accounting DB get updated, this information will be sent to the priority plugin. The ``bulk_update.py`` script can also be configured to run as a cron job to send the updated values to the plugin:
+
+.. code-block:: console
+
+ $ flux account-priority-update -p path/to/DB
```
garlick, line 369: We've got three different ways of specifying the path/to/DB in each script (positional, -f, and -p).  Is that a doc issue or a user interface issue?  Might want to fix that in the code if it really is that way.  Is there a default path to the database built into flux-accounting?  If the default is reasonable maybe we could just state up front what it is and leave off the options in the doc.

adminguide.rst
```diff
@@ -320,6 +322,52 @@ job queue and record of past jobs.
...
+
+.. code-block:: toml
+
+ [job-manager]
+ plugins = [
+   { load = "mf_priority.so"
+   },
+ ]
+
+After the DB and job priority plugin are set up, the ``update-usage`` subcommand should be set up to run as a cron job. This subcommand fetches the most recent job records for every user in the flux-accounting DB and calculates a new job usage value. This subcommand takes one optional argument, ``--priority-decay-half-life``, which, like the parameter set in the database creation step above, represents the number of weeks to hold one job usage "chunk." If not specified, this optional argument also defaults to 1 week.
```
garlick, line 353: Should the first runs of these commands be part of instructions for initializing the database?

Sys admins will want to know the userid and frequency for cron setup.  And it sounds like there are ordering requirements?  If all the scripts should run with the same frequency one after the other, consider showing an example shell script that runs them in the appropriate order.

garlick (changes requested):
Nice work @cmoussa1.  I have a few suggestions, hopefully not too nitpicky.

adminguide.rst
garlick, line 329: Also, would be good to end lines at 80 columns rather than wrapping,  for consistency with the rest of the document, and to be better able to review changes in diff form (diff is line oriented).

adminguide.rst
garlick: BTW, see storage config above.  We've already set up `/var/lib/flux` to be owned by user flux and have plenty of space.  Maybe that is where the db should be?   We could replace /path/to/db with that in the provided commands for easy cut and paste.

adminguide.rst
cmoussa1: Okay! I like this suggestion, thank you - I can make this change.

> the job manager configured to automatically load the multifactor priority plugin provided by `flux-accounting`.

I didn't know this was possible! Does the job manager automatically load the multifactor priority plugin provided by `flux-accounting` via TOML file?

garlick: Yes that is the purpose of the TOML config.

adminguide.rst
cmoussa1: > Is there no other database initialization required? Add users? Associate banks to users?

Other than specifying those two optional parameters, I don't believe there is any other database initialization required prior to adding banks and users. The command just below this text is enough to create a flux-accounting DB. 

Perhaps it would be helpful to add a small subsection that provides an example on adding a couple of banks and then adding a couple users to those banks? Along with adding those subsections like you mentioned?

adminguide.rst
cmoussa1: I'm sorry in advance if this is a dumb answer, as I'm still new to cluster management and whatnot, but I _believe_ the database should be located on the management node - is this the node designated as the flux rank 0 node? The only storage requirements that I can think of is just enough space to hold a sqlite file that could holds all of the flux-accounting information - users, banks, shares, etc.

> We've already set up /var/lib/flux to be owned by user flux and have plenty of space. Maybe that is where the db should be? We could replace /path/to/db with that in the provided commands for easy cut and paste.

I like that suggestion. I can make that change to that example command.

adminguide.rst
garlick: It's the adding banks and users that I thought maybe should be described here?  Or should that be in another document?

What happens if the admin follows all the instructions here, but doesn't add users and banks?   Can the system run jobs?

adminguide.rst
garlick: Yes, the rank 0 node is how we referred to it in the storage section.   So for consistency you could use the same language here and put management node in parenthesis if you like :-)

adminguide.rst
cmoussa1: That would be a user interface issue. Thanks for pointing this out. I will open an issue on flux-accounting to unify the path specification to the flux-accounting DB.

> Is there a default path to the database built into flux-accounting?

The default path to the database should be `/var/FluxAccounting.db`. Stephen had set that up when he added autotools support to the flux-accounting project.

garlick: Maybe we should change it to `@X_LOCALSTATEDIR@/lib/flux/FluxAccounting.db` for your 0.11 release?  Then scripts would just work for the system instance packages without requiring options.

garlick: If these commands should run be run as the flux user, then that should be specified as well.

cmoussa1:
OK, I just pushed an additional commit that attempts to address the great feedback/suggestions I got. Let me know what you think and if it looks any better, and if it does, I can squash it down.

garlick:
I did notice that one of the commands still requires a path to the database?  Does that need to be fixed too?

I think you can state at the top the create database section that you will be creating  the database at the default location: `/var/lib/flux/FluxAccounting.db` and that all commands should be run as the flux user (since earlier in the admin guide we said to create that directory so it is writable by the flux user).

For consistency with the rest of the doc, the TOML fragment should include a full path and be named for the top level key, e.g.  `/etc/flux/system/conf.d/job-manager.toml` (see the format used in sections above)

The automatic updates section should definitely include a suggested frequency (hourly?  daily?), and should state that the scripts must run as the flux user.

If you want to make those changes and then squash, I can make one final pass and then this can go in.  Nice work!

cmoussa1:
Thanks for taking the time to review @garlick! I hope I addressed your suggestions to your liking.

> I did notice that one of the commands still requires a path to the database? Does that need to be fixed too?

That path in that command is in fact not to the flux-accounting database, but to the job-archive database, where it fetches job records. Does that DB have a default location? Is it also in `/var/lib/flux/`?

garlick:
> That path in that command is in fact not to the flux-accounting database, but to the job-archive database, where it fetches job records. Does that DB have a default location? Is it also in /var/lib/flux/?

Ohhhh, sorry!  In fact, that module isn't set up by default, so that's a step we missed.    Let's assume it uses the same path.  We can follow up with another PR adding any config instructions after we review that.

cmoussa1:
Gotcha! In that case, I can squash up a fix that removes that path from the sample code block. :-)

garlick:
It's fine either way.  I'll be following up with the job-archive config stuff so I can retweak it then.

cmoussa1:
OK, sounds good. And I can open an issue in flux-accounting to at least track adding default path support for `flux account update-usage` in the meantime.

garlick (approved):
Thanks for all the work @cmoussa1 !  This looks good for now.
As I said I'll probably be going through and polishing the document as a whole tomrorow, as well as adding config for job-archive (which it looks like we are not loading in rc1 so that needs to get fixed in flux-core!)  So hopefully you'll be OK if I tweak things a bit then to improve the overall continuity?

I'll probably also try to walk through the instructions one more time and see if I find any missing steps that way.

cmoussa1:
> As I said I'll probably be going through and polishing the document as a whole tomrorow, as well as adding config for job-archive (which it looks like we are not loading in rc1 so that needs to get fixed in flux-core!) So hopefully you'll be OK if I tweak things a bit then to improve the overall continuity?
>
> I'll probably also try to walk through the instructions one more time and see if I find any missing steps that way.

I think that sounds like a great plan. Feel free to ping me if you find any other room for improvement by our release date tomorrow.

garlick:
Sounds good!  Want to set MWP?

cmoussa1:
Sure. Will do.

PR #114, opened 2021-11-04
faqs: update sqlite redirection instructions + misc cleanup

garlick:
Problem: the FAQ recommends setting _rundir_ to relocate the sqlite database file, but this can cause problems with UNIX domain sockets as noted in flux-framework/flux-core#3925.

Update this FAQ entry and add some other information that might prove helpful.

Plus some other cleanup in this rst file.


chu11 (approved):
LGTM

PR #115, opened 2021-11-05
adminguide: document job-manager prolog and epilog

grondo:
This is a first draft of a section for the Administrator's Guide describing how to configure job-manager prolog and epilog, and how to use the included `flux perilog-run` command to run a per-rank prolog/epilog as root if required.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -439,6 +439,82 @@ following example:
  flux account-update-fshare
  flux account-priority-update
 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Job prolog/epilog configuration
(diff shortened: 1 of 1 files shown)


garlick (approved):
Let's put this in as is for now, with the caveat that we may want to consolidate job manager configuration in one TOML file rather than require an admin following the guide to piece this fragment and the one in the accounting section together.  I was planning on making a continuity pass on the doc later today anyway and can address that, or at least try it and see if it helps.

garlick:
I'm just going to go ahead and set MWP here - hope that's OK!

grondo:
I was having the same thought. We need a configuration reference manual in
addition I think.

PR #117, opened 2021-11-11
Add short FAQ entry for `flux mini bulksubmit` "hangs"

grondo:
Problem: Users just starting with `flux mini bulksubmit` may be
confused that it waits on stdin by default, apparently hanging.

Add a FAQ entry to which we can point users or they perhaps discover
on their own.

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -197,3 +197,40 @@ A special `job shell plugin <https://github.com/flux-framework/flux-pmix>`_,
 offered as a separate package, is required to bootstrap the upcoming openmpi
 5.0.x releases.  Once installed, the plugin is activated by submitting a job
 with the ``-ompi=openmpi@5`` option.
+
+---------------------------------------------------
+Why does the ``flux mini bulksubmit`` command hang?
+---------------------------------------------------
+
+The ``flux mini bulksubmit`` command works similar to GNU parallel or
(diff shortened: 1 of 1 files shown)


faqs.rst
```diff
@@ -197,3 +197,32 @@ A special `job shell plugin <https://github.com/flux-framework/flux-pmix>`_,
...
+
+The ``flux mini bulksubmit`` command works similar to GNU parallel or
+``xargs`` and is likely blocked waiting for input from ``stdin``.
+Typical usage is to send output of some command to ``bulksubmit`` and,
+like ``xargs -I``, substitute the input with ``{}``. For example:
+
+.. code-block:: console
+
+ $ seq 1 4 | flux mini submit echo {}
+
```
chu11, line 213: suggestion, also say what this does or show the output result.

grondo: ha, I even see a typo (`submit` instead of `bulksubmit`), so thanks for calling this out. Maybe we should use a more useful example as well

grondo: Thanks! I added the full output (including the `--watch` option so that output is copied to the terminal). I couldn't think of a better example suitable for a FAQ, so I left the command as is.

chu11 (approved):
LGTM!

grondo:
Thanks!

PR #120, opened 2022-01-05
update mergify config and small admin guide fixes

grondo:
This is a small collection of fixes:

 * update mergify config to use queues since strict merge is no longer supported
 * Fix an incorrect path to imp config in admin guide
 * update output of `flux overlay status` in admin guide (this one may be premature, but I went ahead and updated output while I was thinking about it) Happy to drop this one for now.


garlick (approved):
Thanks!

PR #121, opened 2022-01-13
adminguide: update flux-core version references, clarify prolog/epilog documentation

grondo:
This PR updates flux-core version refs from 0.32.0 to 0.33.0.

Also, it turns out the job prolog/epilog section was pretty confusing, so the description has been expanded somewhat in hopes of alleviating any confusion between job vs job-manager prolog/epilog,  IMP "run" support, and the `/etc/flux/system/{prolog,epilog}.d` directory. (Unsure if it is any clearer now)

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -454,15 +454,22 @@ And the scripts should be run by ``flux cron``:
 Job prolog/epilog
 =================
 
-.. warning::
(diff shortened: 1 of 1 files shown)


garlick (approved):
Thanks!  Good improvement.

PR #122, opened 2022-01-15
adminguide: drop manual install instructions

garlick:
Problem: the admin guide gives instructions for manually installing
a Flux system instance, but this is unlikely to be an acceptable
practice at most sites and just decreases the signal to noise ratio
in the document.

Drop the installation description and add a 'note' box directing
would-be Flux package maintainers to open an issue for assistance.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -115,7 +115,8 @@ A system user named ``flux`` is required, with the following characteristics:
 Software
 ========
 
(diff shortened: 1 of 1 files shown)


grondo:
This LGTM and is a nice improvement.

I now notice that mergify is marking the configuration as invalid. I'll merge this manually and fix the mergify config in a future PR.

grondo (approved):

PR #123, opened 2022-01-18
mergify: fix mergify config

grondo:
Problem: The .mergify.yml had a line that was off by one space,
and thus could not be parsed as YAML.

Fix the invlalid YAML, also move some conditions back to
pull_request_rules.


garlick (approved):
thanks

PR #127, opened 2022-01-26
adminguide: fix bootstrap example

garlick:
As noted in flux-framework/flux-core#4067, the admin guide `[bootstrap]` example doesn't match the text.  Fix that.


grondo (approved):
LGTM

PR #128, opened 2022-01-31
add link, refs, and intersphinx config for flux-security

grondo:
This PR adds a link to the new flux-security man pages, as well other configuration bits to make reference flux-security docs possible in this high-level documentation repo.


garlick (approved):
LGTM!  I duplicated some of this in #129 so when this goes in I'll rebase that one so it only updates the admin guide.

grondo:
Sounds good, looks like I missed some of the other spots you added references as well.

BTW, I had to configure flux-security as a subproject of flux-framework in readthedocs as well to get this to work.

PR #132, opened 2022-02-16
WIP: correct accounting setup process

garlick:
Feel free to push any accounting-related tweaks directly to this PR as @ryanday36 works through flux accounting setup on fluke.


garlick:
Oops, apparently one of the fixes here was submitted twice.   The other one can be combined with the PR to update for the next flux-core release, so withdrawing this one.

PR #134, opened 2022-02-25
adminguide: fix [archive] TOML example

garlick:
Problem: the [archive] configuration example uses integer seconds where FSD is required.

Update example.

This was noted in flux-framework/flux-core#4163.


grondo (approved):
LGTM!

PR #136, opened 2022-03-31
adminguide: updates for v0.38.0

grondo:
This is a WIP which adds documentation of a couple new features in the flux-core v0.38.0 release.

Just parking this here for now in case someone wants to correct, amend, emend, or expand on these two new sections thus far.

I guess we should also update all the "notes" to reference the new release as well.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -13,10 +13,10 @@ resource manager on a cluster.
     in this guide may change with regularity.
 
     This document is in DRAFT form and currently applies to flux-core
-    version 0.36.0.
+    version 0.38.0.
 
 .. warning::
-    0.36.0 limitation: the flux system instance is primarily tested on
+    0.38.0 limitation: the flux system instance is primarily tested on
     a 128 node cluster.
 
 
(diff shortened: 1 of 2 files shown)


garlick:
I pushed a few more updates - hope that's ok.

grondo:
Yes, that was what I was hoping! I'll take a look in the morning

adminguide.rst
```diff
@@ -498,6 +498,8 @@ TOML table, as shown in the example below:
 
 .. code-block:: toml
 
+  # /etc/flux/system/conf.d/ingest.toml
```
grondo, line 501: This seems fine, but I was wondering if our implication that the toml config for every subsystem has to be in a different file is going to drive sysadmins crazy. Is there some benefit to implying that?

garlick: Yeah good point.   I was just going for consistency here, but I'm open to changing the overall presentation.

grondo: Yeah, I was more thinking out loud at this point

grondo:
Agreed. Setting MWP.

garlick:
@mergify-io refresh

PR #138, opened 2022-04-03
quickstart: remove references to flux-hwloc

garlick:
Problem: the `flux-hwloc` command is being removed (flux-framework/flux-core#4262) but this command is referred to several times in the quickstart guide.

Substitute other commands.

### DIFF
--- a/quickstart.rst
+++ b/quickstart.rst
@@ -184,7 +184,7 @@ To start a Flux instance with 4 brokers on the local node, use ``flux start``:
 
 A Flux instance can be also be started under `Slurm
 <https://github.com/chaos/slurm>`_ using PMI. To start by using ``srun(1)``,
-simply run the ``flux start`` command without the ``--size`` option under
+simply run the ``flux start`` command without the ``--test-size`` option under
 a Slurm job. You will likely want to start a single broker process per node:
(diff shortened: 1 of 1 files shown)


quickstart.rst
```diff
@@ -234,12 +234,14 @@ Interacting with a Flux Session
 
 There are several low-level commands of interest to interact with a Flux
 instance. For example, to view the total resources available in the current
-instance, ``flux hwloc info`` may be used:
+instance, ``flux resource status`` may be used:
```
grondo, line 237: `flux resource status` doesn't give you the total resources available in the instance (just the status of the execution targets). Maybe change to "to view the status of resources in the current instance", or replace with `flux resource list` and drop the "To view the scheduling state of resources".

This might actually be a good place to introduce `flux uptime` or `flux top` also.

grondo (approved):
LGTM, though I had one comment inline.

quickstart.rst
grondo: Hm, also, these aren't exactly low-level commands anymore. I wonder if instead of worrying about this now, we should merge this PR as is and come back around to "refresh" this entire document.

PR #137, opened 2022-03-31
fluxion: drop use of flux-ion-R to augment the system R

dongahn:
Problem: flux-ion-R augments the R encoded by flux-R for our fluxion
scheduler. However, the full R with the scheduling key (with which we
have a system resource graph) is difficult to manage and debug at the
system-instance level. To fix this problem, we just added rv1exec reader
support directly into the fluxion scheduler (flux-sched PR #9210)
and we now need our documentation update accordingly.

Remove the prose on flux-ion-R from the admin guide section.

PR #139, opened 2022-04-05
adminguide: use single TOML files for examples

garlick:
This PR switches the admin config file examples over to using one big example file (for each of the three config spaces) rather than a lot of little files that might be annoying to sys admins.

The examples are not all-inclusive.  accounting, prolog/epilog, and validator configs are still presented in separate "add on" sections.  I did do some re-titling of sections and tweaked the header level in some cases.

Maybe give the rendered version a read and see if it's any better?   I'm not hard over on this change, but I do think it might be a little better than before.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -141,35 +141,38 @@ flux-accounting
     distribution so we can share packaging tips and avoid duplicating effort.
 
 
-*******************
-Configuration files
-*******************
+*************
+Configuration
+*************
 
 Much of Flux configuration occurs via
-`TOML <https://github.com/toml-lang/toml>`_ configuration files found
-in a hierarchy under ``/etc/flux``.  For the most part, Flux
-components will read all TOML files from a given directory in a glob,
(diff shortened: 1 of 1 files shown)


garlick:
I just realized this includes the proposed change in #137 to remove the note about `flux ion-R`.  I can fix that if needed.

grondo (approved):
Wow, this reads _much_ better in my opinion! I also think admins will appreciate seeing a large chunk of configuration all in one place, rather than piecemeal in the document. 

This was a large amount of work, thanks!

grondo:
>  I just realized this includes the proposed change in #137 to remove the note about flux ion-R. I can fix that if needed.

Hm, seems like we could just merge as is and drop the other PR if that's more efficient (it is just documentation anyway)

garlick:
Thanks!   I just fixed a typo and re-pushed.  Well based on that glowing review, maybe I'll just set MWP then :smile:

PR #141, opened 2022-04-21
faq: update content.sqlite instructions

garlick:
Problem: instructions for relocating content.sqlite refer to
'content.backing-path', which is no longer a valid broker attribute.

Update the instructions to suggest setting 'statedir' and reword
instructions to be more clear now that the database file path itself
cannot be explicitly set.

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -64,26 +64,27 @@ Memory exhaustion on a node when running large ensembles with Flux
 
 Flux's in-memory KVS, or more properly, its content-addressable storage
 subsystem, is backed by an `SQLite <https://www.sqlite.org>`_ database file,
-located by default in ``/tmp`` (or ``$TMPDIR``).  On some systems, ``/tmp``
-is a RAM-backed file system.  To minimize Flux's memory footprint on such
-systems, Flux may be launched with the database file redirected to a more
-appropriate location.  For example:
+located by default in *rundir*, which is usually in ``/tmp``.  On some systems,
+``/tmp`` is a RAM-backed file system.  To minimize Flux's memory footprint
+on such systems, Flux may be launched with the database file redirected to
+a more appropriate location by setting the *statedir* broker attribute.  For
+example:
 
 .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


faqs.rst
```diff
@@ -64,26 +64,27 @@ Memory exhaustion on a node when running large ensembles with Flux
...
 subsystem, is backed by an `SQLite <https://www.sqlite.org>`_ database file,
-located by default in ``/tmp`` (or ``$TMPDIR``).  On some systems, ``/tmp``
-is a RAM-backed file system.  To minimize Flux's memory footprint on such
-systems, Flux may be launched with the database file redirected to a more
-appropriate location.  For example:
+located by default in *rundir*, which is usually in ``/tmp``.  On some systems,
+``/tmp`` is a RAM-backed file system.  To minimize Flux's memory footprint
+on such systems, Flux may be launched with the database file redirected to
+a more appropriate location by setting the *statedir* broker attribute.  For
+example:
```
grondo, line 71: Though the current wording makes sense, I keep worrying that the implicit statement that `statedir` overrides `rundir` for the database file should be made explicit. However, I'm having a hard time crafting alternative wording that isn't more clumsy than what you have here. This is the best I could come up with, take or leave it, it probably isn't a major concern.

```suggestion
located in *rundir* by default unless the *statedir* broker attribute is set.
Because *rundir*, which is usually in ``/tmp``, may point to a RAM-backed
filesystem, it may be advisable in some cases to launch Flux with the database
in a more appropriate location by setting *statedir*.
example:
```

grondo (approved):
LGTM! Thanks! 

I thought that it should somehow be explicitly stated that the database file is located in `rundir` _unless_ `statedir` is set (but that could just be me). I tried to craft alternative wording, but unfortunately my prose seems more confusing somehow so feel free to merge as is.

faqs.rst
garlick: Hmm, I'm not having any luck coming up with an alternative wording either.  There is a link to the longer description in flux-broker-attributes(7) so I suppose people can easily dig a little deeper.  I'd say let's merge this and we can always submit another PR if better wording comes to mind.

Issue #70, opened 2020-10-02
Using `flux-proxy` on a machine with launch nodes


SteVwonder:
For Summit (and Lassen, depending on if you use lalloc or not), the default behavior is to allocate compute nodes and a launch node for every job.  The batch script is then run on the launch node and any use of the system parallel launcher `jsrun` to execute Flux results in all of the Flux processes to land on the compute nodes and none on the launch node.  If a user captures the `local://` FLUX URI and tries to use that from the launch node, it will fail, since the Unix socket is on the compute nodes not the launch node.  This behavior is different from typical clusters where launch nodes are not used.

SteVwonder:
One potential solution is to use to construct and use an `ssh://` FLUX URI to leverage the ssh connector rather than the local connector.  This will create an ssh connection between the launch node and the compute node on which Flux is running.

garlick:
That should be doable, provided you can somehow capture FLUX_URI from one of the brokers and transform it.   Alternate idea would be to set the `local-uri` attribute on the broker command line to something you know in advance, but careful - if command line applies to all brokers, it has to work for all.

Does jsrun have options by chance to include the launch node in the "job"?  If you could locate flux rank 0 on the launch node, and then exclude it from scheduling, that would be sorta convenient.  The initial program / shell would then run on the launch node and the system's model would be preserved.

How do we usually run on this system?

SteVwonder:
> Does jsrun have options by chance to include the launch node in the "job"? 

That's a good question. I'm not sure, but I think not.  I don't see one documented in [LLNL](https://hpc.llnl.gov/banks-jobs/running-jobs/slurm-srun-versus-ibm-csm-jsrun) or [ORNL's docs](https://docs.olcf.ornl.gov/systems/summit_user_guide.html#node-types).  I did see this in the ORNL ones though:

> All commands within your job script (or the commands you run in an interactive job) will run on a launch node. Like login nodes, these are shared resources so you should not run multiprocessor/threaded programs on Launch nodes.

If there is a way, I suspect running rank 0 of a large Flux instance on the Launch node would be frowned upon due to all the activity the KVS and other rank 0 modules generates.

> How do we usually run on this system?

At LLNL, we typically use `lrun` (and its variants) which by default runs your job script on a compute node, so this particular issue isn't common on LLNL systems.  ORNL does not have a similar wrapper script or setup.

[...]

PR #146, opened 2022-07-25
docs: add information on running Flux interactively on CORAL

wihobbs:
Now that we have the LSF URI resolver, we should add some documentation so users can take advantage of its utility.

We should probably wait until we release & tag flux-core@0.42.0 and get that built on BlueOS to merge this.

### DIFF
--- a/coral.rst
+++ b/coral.rst
@@ -128,3 +128,59 @@ each binding to 10 CPU cores and 1 GPU on a compute node:
 .. code-block:: sh
 
   flux mini run -N 1 -n 4 -c 10 -g 1 -o mpi=spectrum -o cpu-affinity=per-task -o gpu-affinity=per-task my_mpi_binary
+
+----------------------------
+Launching Flux Interactively
+----------------------------
+
+CORAL systems use IBM's Spectrum LSF scheduler which, unlike Slurm, does not support running ``jsrun`` commands in a pseudo-
+terminal. This limits users' ability to run Flux sessions interactively.
+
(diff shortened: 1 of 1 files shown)


wihobbs:
Updating commit message -- just looked through our issues and I believe this will close #70.

garlick:
Nicely done!  And thanks for finding that issue.

garlick (approved):
Oops this should have been merged a while back!

wihobbs:
The `make-check` issue should be fixed but the `make-linkcheck` is failing because LC confluence is behind an auth wall.

grondo:
Thanks @wihobbs! sphinx has a [`linkcheck_ignore`](https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=linkcheck#confval-linkcheck_ignore) setting we can use to ignore all links in LC confluence if you want to try that?

wihobbs:
Hi @grondo! Will do.

wihobbs:
I put some more time into this and I couldn't seem to get `linkcheck_ignore` to work; however, I found that the direct jump to "RZAnsel users should verify SSH keys" (after the hashtag in the link) was causing the problem, so I just removed that. If a user is able to navigate to the quickstart guide, they should be able to find this pretty easily anyway. Fixed and pushed.

Someone should probably click the link and make sure it still works just as a sanity check (I don't have access to LC confluence anymore...)

grondo:
Thanks I approved the workflow to run and looks like it succeeded! :+1:

wihobbs:
Thanks @grondo! and tysm @jameshcorbett for the fix-up :)

PR #148, opened 2022-08-03
adminguide: fix formatting error

garlick:
Problem: example TOML config is improperly formatted.

Insert a leading space where one was missing.


grondo (approved):
Good catch!

garlick:
Thanks!  Sheesh I should have checked that.

PR #149, opened 2022-08-09
docs: add a flux-accounting section

cmoussa1:
#### Background

As the flux-accounting project continues to mature and become more of a part of system-instance Flux, it would be nice to have more documentation about the specifics of flux-accounting, such as database administration, job usage and fair-share information, the multi-factor priority plugin, and more. It might also be nice if this documentation was easy to locate alongside the other Flux components.

---

This is a small [WIP] PR that adds a separate page in flux-docs that adds more details about the flux-accounting sub-project. For the purposes of keeping this PR somewhat simple (and hopefully easy to review!), I've only added one sub-section: a sub-section that outlines basic database administration and a glossary. 

I've written this guide with the intended audience of an administrator (someone who is going to be using flux-accounting), but if perhaps a change of direction would be better, let me know. :-)

I am very much open to any and all feedback on how to make this documentation more clear. For example, if specific database administration examples would help (e.g showing how to use the `add-*` or `edit-*` commands, I would be more than happy to add those.

In the near future, I plan on submitting subsequent PR's that add sections to this page on some of the other major components of flux-accounting, which may include:

- job usage and fairshare documentation
  - examples of how usage is calculated
- multi-factor priority plugin
  - what factors are used, how they affect job priority calculation
  - accounting limits (`max_active_jobs`, `max_running_jobs`)

### DIFF
--- a/flux-accounting-guide.rst
+++ b/flux-accounting-guide.rst
@@ -1,0 +1,95 @@
+.. _flux-accounting-guide:
+
+#####################
+Flux-Accounting Guide
+#####################
+
+*key terms: association, bank*
+
+The *Flux Accounting Administrator's Guide* documents relevant information for
+the administration and management of flux-accounting components alongside a
+system instance install of Flux.
+
+.. note::
+    flux-accounting is still beta software and many of the interfaces 
+    documented in this guide may change with regularity.
+
+    This document is in DRAFT form and currently applies to flux-accounting
+    version 0.18.1.
+
+***********************
+Database Administration
+***********************
+
+The flux-accounting database is a SQLite database which stores user account
+information and bank information. Administrators can add, disable, edit, and
+view user and bank information by interfacing with the database through
+front-end commands provided by flux-accounting. The information in this
+database works with flux-core to calculate job priorities submitted by users,
+enforce basic job accounting limits, and calculate fair-share values for
+users based on previous job usage.
+
+Each user belongs to at least one bank. This user/bank combination is known 
+as an *association*, and henceforth will be referred to as an *association* 
+throughout the rest of this document.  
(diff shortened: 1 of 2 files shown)


garlick:
This is excellent!  I had two quick thoughts:
- consider adding a glossary for terms like "bank" and "association"
- should we consider moving the accounting section in the flux admin guide to here and leaving a reference there?

cmoussa1:
> consider adding a glossary for terms like "bank" and "association"

Yes, I like this suggestion, thank you! I've already added a *glossary* for _association_, but it would definitely be worth adding a definition for a bank too (along with all the other accounting-specific terms once they are added to this document). I'll add a _bank_ definition before I mark this as ready for review.

> should we consider moving the accounting section in the flux admin guide to here and leaving a reference there?

I like this suggestion too - I wonder, though, if it might be worth leaving the basic installation/configuration instructions for flux-accounting in the flux admin guide, and moving any nitty gritty details (like database administration, for example) to this document. :-) Just my opinion though. Does that make sense and would be okay with you?

garlick:
Oh duh!  In my quick skim I must have registered your glossary subliminally and then it became my idea!  Wow.  Brain not working right.

Sure, I'm ok with leaving the admin guide stuff where it is.   The two guides can be complementary.

cmoussa1:
Lol, no problem! Glad we are on the same page! I've just pushed an additional definition of a _bank_, and I'll go ahead and mark this as ready for some review whenever convenient. Thanks for your review and suggestions so far.

garlick (approved):
Great start!  LGTM!

cmoussa1:
Thanks @garlick! Setting MWP

PR #150, opened 2022-08-11
[WIP] docs: add job usage calculation section

cmoussa1:
This is a [WIP] PR that adds a new section to the flux-accounting Administrator's Guide that outlines how it performs current and historical job usage calculation for associations.

It also adds the key terms used in this section, **PriorityDecayHalfLife** and **PriorityUsageResetPeriod**, to the glossary at the bottom of the page.

(leaving [WIP] for now while I consult the documentation in flux-framework/flux-accounting to make sure I didn't leave out anything important)


cmoussa1:
OK, think I'll mark this as ready for some initial review whenever convenient. :-)

garlick (approved):
Sorry for the delay @cmoussa1!  LGTM.

One thing, maybe for a future PR, is that we've reused a lot of terminology and even variable names from slurm saccounting design.  If we're going to do that, it feels like the documentation should contain a citation, and perhaps point out that the reason the systems are similar is that both slurm and flux accounting systems were developed at LLNL to meet local management requirements, and since the code is open source, we felt that it was appropriate to reuse much of slurm's design in flux.

garlick:
Oh I forgot, I was going to also ask if the intention was to put configuration parameters in the glossary?  It seems like the glossary should define terms, and configuration variables could be defined separately (summarized in another section if needed).

cmoussa1:
> it feels like the documentation should contain a citation, and perhaps point out that the reason the systems are similar is that both slurm and flux accounting systems were developed at LLNL to meet local management requirements, and since the code is open source, we felt that it was appropriate to reuse much of slurm's accounting design in flux.

Completely agree. I really like this suggestion, thanks for pointing this out @garlick! I'll open an issue to add a citation in a near future PR. 

> Oh I forgot, I was going to also ask if the intention was to put configuration parameters in the glossary? It seems like the glossary should define terms, and configuration variables could be defined separately (summarized in another section if needed).

That's a good point - I was including both in the glossary, but as you pointed out, perhaps it would be better to define configuration variables in their own section, since there are at least a couple. I'll force push a fix to just remove those parameter definitions from the Glossary for now.

PR #153, opened 2022-08-15
adminguide: add [ingest] table to example config

garlick:
Problem: the example flux system configuration does not
call the feasibility validator plugin, so jobs with unsatisfiable
resource requests are terminated with an exception rather than
being rejected at submit time.

Add [ingest] to the example.


grondo (approved):
Good catch!

PR #154, opened 2022-08-16
adminguide: add fluxion config

garlick:
Problem: the adminguide recommends installing flux-sched
but omits configuration from the example config file.

Add example and man page references.


grondo (approved):
Nice catch!

PR #155, opened 2022-08-18
faqs: add faq about overcommitting resources

grondo:
After the most recent Flux tutorial, a user asked if there was an equivalent to Slurm's `--overcommit` option, which can be used to run more than one task per allocated core.

It seemed like a good question with multiple answers in Flux (at least currently). So I've added an FAQ entry.


garlick (approved):
Good one, thanks!

PR #156, opened 2022-08-19
faqs: update based on early flux rollout on production systems

garlick:
The FAQ didn't have any categories so I added a General category for most everything we have now, and another for MPI, and added some new entries to the MPI section based on our fun with mvapich and openmpi this past week.

A couple of other entries were updated, and one on F58 encoding issues was added.

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -195,36 +195,6 @@ I'm experiencing a hang while running my parallel application. How can I debug?
 
 .. _versioning_multi_repo:
 
-How does the versioning of Flux work with its multi-repo structure?
-===================================================================
-
-For any given repository, the versioning is typical semantic `versioning <https://semver.org/>`_.
-All of the Flux repos are still < v1.0, so all of our interfaces are subject
-to change. Once a repo hits v1.0, the interfaces for that repo will only break
-backwards compatibility on major version increments. New features get added in
-minor releases. Etc
-
-The interesting part of the versioning comes from the multi-repo structure.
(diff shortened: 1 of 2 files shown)


faqs.rst
```diff
@@ -1,35 +1,60 @@
...
+*****************
+General Questions
+*****************
+
+What's with the fancy ƒ?
+========================
+
+Flux job IDs and their multiple encodings are described in
+`RFC 19 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_19.html>`_. The default F58 encoding assumes that your terminal session can
+handle UTF-8.
```
grondo, line 21: I wonder if we should state something like: The UTF-8 `ƒ` character is used to denote a Flux jobid in F58 encoding when the current locale supports UTF-8 multi-byte characters. When the locale does not support UTF-8, then an ASCII `f` is used. The ASCII rendering of jobids can be forced by setting ``FLUX_F58_FORCE_ASCII=1`` in your environment.

Actually, the main point is that Flux will try to first determine if the current locale supports UTF-8 multi-byte characters before using `ƒ`. Maybe we just say that? Up to you.

grondo (approved):
This is great, thanks!

Just one suggestion inline you can take it or leave it.

faqs.rst
garlick: Great point.  I'll mention something to that effect.

garlick: I forced an update - look OK?

grondo: Yes, nice!

PR #157, opened 2022-08-20
faqs: update job steps entry

garlick:
The FAQ entry titled "How do I mimic Slurm's job step semantics" probably will leave users with more questions than answers.

Take a crack at expanding this explanation and reorienting the curious towards Flux's way of doing things.

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -170,7 +170,7 @@ Note the following:
 * *statedir* must exist before starting Flux.
 * If *statedir* contains ``content.sqlite`` it will be reused.  Unless you are
   intentionally restarting on the same nodes, remove it before starting Flux.
-* Unlike *rundir*, *statedir* and the ``content.sqlite`` file with in it
+* Unlike *rundir*, *statedir* and the ``content.sqlite`` file within it
   are not cleaned up when Flux exits.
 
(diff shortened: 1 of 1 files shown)


faqs.rst
```diff
@@ -170,19 +170,43 @@ Note the following:
...
-===========================================
+How do I run job steps?
+=======================
 
-Using ``flux mini submit`` to submit a script containing multiple
-``flux mini run`` invocations will not result in Slurm-style job steps unless
-the job script is prefixed with ``flux start`` .
+A Flux batch job or allocation started with ``flux mini batch`` or
+``flux mini alloc`` is actually a full featured Flux instance run as a job
+within the Flux system instance.  Unlike SLURM, Flux does not have a separate
```
grondo, line 185: I wonder if this might confuse readers that `flux mini batch` and `flux mini alloc` can only be used in a system instance, though of course that is the common case, so perhaps it doesn't matter?

If anything, maybe just substitute "enclosing instance" in place of "system instance"?

Not sure if that would make it more confusing though

garlick: Sounds good - just force pushed that change.

grondo (approved):
Nice improvement!

PR #158, opened 2022-08-20
faqs: rework "launch large number of jobs" entry

garlick:
Problem: the "how do I efficiently launch a large number of jobs"
FAQ entry refers to the outdated workflow examples, and doesn't
provide a lot of useful info.

Expand this section, and add the now important suggestion to
run jobs in a sub-instance, not the system instance.

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -71,16 +71,37 @@ If no output is produced, then your hwloc is not CUDA-enabled.
 
 .. _launch_large_num_jobs:
 
-How do I efficiently launch a large number of jobs to Flux?
-===========================================================
+How do I efficiently launch a large number of jobs?
+===================================================
 
-See `bulksubmit.py <https://github.com/flux-framework/flux-workflow-examples/tree/master/async-bulk-job-submit>`_
-for an example workflow. You can also submit many copies of the same job using
-``flux mini submit --cc=IDSET``, or submit many jobs based on a series of
-inputs on ``stdin`` or the command line via ``flux mini bulksubmit``. See the
-`flux-mini <https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-mini.html#bulksubmit>`_
-manual page for more details.
+If you have many jobs to run here are some tips to help improve efficiency
+and throughput:
 
+- Create a batch job or allocation to contain the jobs in a Flux sub-instance.
+  This improves performance over submitting them directly to the Flux system
+  instance and reduces the impact of your jobs on system resources and other
+  users.  See also: :ref:`batch`.
+- If scripting ``flux mini submit`` commands, avoid the pattern of one command
+  per job as each command invocation has a startup cost.  Instead try to
+  combine similar job submissions with ``flux mini submit --cc=IDSET``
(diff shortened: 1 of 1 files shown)


faqs.rst
```diff
@@ -71,16 +71,31 @@ If no output is produced, then your hwloc is not CUDA-enabled.
...
+  users.  See also: :ref:`batch`.
+- An obvious one, but don't wait for the first job to finish before submitting
+  the next job.  For example, replace ``flux mini run`` with
+  ``flux mini submit``.  The ``flux queue drain`` command may be used to wait
+  for all jobs to complete.  See also: :core:man1:`flux-mini`,
+  :core:man1:`flux-queue`.
+- Combine similar job submissions together using ``flux mini submit --cc=IDSET``
+  or `flux-mini builksubmit <https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-mini.html#bulksubmit>`_.
+  Each invocation of the command line tool has a start-up cost, so combining
+  them is helpful, where possible.
```
grondo, line 92: Maybe an added note here:

> By default `flux mini submit --cc=IDSET` and `flux mini bulksubmit` will exit once all jobs have been submitted. To wait for all jobs to complete before proceeding, use the `--wait` or `--watch` options to these tools, or use `flux queue drain` or `--flags=waitable` as described above..

faqs.rst
```diff
@@ -71,16 +71,31 @@ If no output is produced, then your hwloc is not CUDA-enabled.
...
+
+- Create a batch job or allocation to contain the jobs in a Flux sub-instance.
+  This improves performance over submitting them directly to the Flux system
+  instance and reduces the impact of your job on system resources and other
+  users.  See also: :ref:`batch`.
+- An obvious one, but don't wait for the first job to finish before submitting
+  the next job.  For example, replace ``flux mini run`` with
+  ``flux mini submit``.  The ``flux queue drain`` command may be used to wait
+  for all jobs to complete.  See also: :core:man1:`flux-mini`,
+  :core:man1:`flux-queue`.
```
grondo, line 88: What about shortening to:

> The `flux mini run` command will block until the job completes. Use `flux mini submit` instead to submit multiple jobs to be scheduled and run in the background, then wait for all queued jobs to complete with `flux queue drain`. Or use the `--flags=waitable` option to `flux mini submit` and use `flux job wait --all` (this has the benefit of reporting failed jobs and exits with non-zero exit status if any job fails).

Ok, that isn't shorter, but perhaps the note about exit status is useful?

grondo (approved):
LGTM! I have just proposed a couple alternate descriptions, but  completely optional!

I was also wondering if we should more strongly state that submitting anything more than a handful of jobs with back-to-back `flux mini submit` calls should be avoided. The question _is_ about efficiently submitting "a large number of jobs", not necessarily how to submit and run more than one job at a time. (Maybe consider rewording the question?)

Eh, these were just questions that came up while reading through the new answers. The new content is definitely helpful.

faqs.rst
garlick: OK, I tried to rework with your suggestions in mind and force pushed.  Could you have another look and let me know what you think?  I'm happy to iterate further if it's not quite there yet.

grondo: Yeah that seems more complete to me, though the original wasn't bad either!

grondo (approved):
Great!

PR #159, opened 2022-08-21
faqs: split missing resources question into two questoins

garlick:
Problem: the "Why is Flux not discovering and managing all of the resources
on the system/node?" question is overly broad and doesn't explain Flux's
behavior very well.

Focus the question on missing nvidia GPUs and explain in more detail how
that problem can manifest itself.

Drop the bit about "srun --mpibind=off" and "jsrun --bind=none" since
that appears to be answering a different question?

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -50,24 +50,42 @@ or `sched <https://github.com/flux-framework/flux-sched/issues>`_.
 
 .. _not_managing_all_resources:
 
-Why is Flux not discovering and managing all of the resources on the system/node?
-=================================================================================
+Why is Flux ignoring my Nvidia GPUs?
+====================================
 
-This can be due to various bind flags that need to be passed parallel launcher
-that started Flux. For example at LLNL you must pass ``--mpibind=off`` to
-``srun`` and ``--bind=none`` to ``jsrun``.
-
-Also on all systems, Flux relies on hwloc to auto-detect the on-node resources
-available for scheduling.  The hwloc that Flux is linked against must be
-configured with ``--enable-cuda`` for Flux to be able to detect Nvidia GPUs.
-
-You can test to see if your system default hwloc is CUDA-enabled with:
+When Flux is launched via a foreign resource manager like SLURM or LSF,
+it must discover available resources from scratch using
+`hwloc <https://www.open-mpi.org/projects/hwloc/>`_.  To print a resource
+summary, run:
 
 .. code-block:: sh
 
-  lstopo | grep CoProc
+  $ flux resource info
+  16 Nodes, 96 Cores, 16 GPUs
 
-If no output is produced, then your hwloc is not CUDA-enabled.
+The version of hwloc that Flux is using at runtime must have been configured
+with ``--enable-cuda`` for it to be able to detect Nvidia GPUs.  You can test
+to see if hwloc is able to detect installed GPUs with:
+
+.. code-block:: sh
+
+  $ lstopo | grep CoProc
+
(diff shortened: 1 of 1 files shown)


faqs.rst
```diff
@@ -50,24 +50,38 @@ or `sched <https://github.com/flux-framework/flux-sched/issues>`_.
...
 .. _not_managing_all_resources:
 
-Why is Flux not discovering and managing all of the resources on the system/node?
-=================================================================================
+Why is Flux ignoring my Nvidia GPUs?
+====================================
 
-This can be due to various bind flags that need to be passed parallel launcher
-that started Flux. For example at LLNL you must pass ``--mpibind=off`` to
-``srun`` and ``--bind=none`` to ``jsrun``.
```
grondo, line 58: I think there was indeed an issue where default mpibind and possibly `jsrun` binding made it so that hwloc couldn't detect GPUs (maybe they set `CUDA_VISIBLE_DEVICES` to something incorrect by default?)

If we haven't verified that these options are no longer needed, perhaps we leave this in as something to try, e.g.

> If running under Slurm or LSF, try disabling any CPU affinity options, e.g. `--mpibind=off` with Slurm or `--bind=none` with `jsrun`.

faqs.rst
```diff
@@ -50,24 +50,38 @@ or `sched <https://github.com/flux-framework/flux-sched/issues>`_.
...
-Also on all systems, Flux relies on hwloc to auto-detect the on-node resources
-available for scheduling.  The hwloc that Flux is linked against must be
-configured with ``--enable-cuda`` for Flux to be able to detect Nvidia GPUs.
+.. code-block:: sh
+
+  $ flux resource list
+       STATE NNODES   NCORES    NGPUS NODELIST
+        free     16       96       16 test[0-15]
+   allocated      0        0        0
+        down      0        0        0
```
grondo, line 67: Suggestion:  We do support `flux resource info`, and its output is more compact.

grondo (approved):
LGTM, a couple optional comments inline (meant to do a review, but accidentally left individual comments)

faqs.rst
grondo: Actually, now that I think about this more, I'm not sure the binding options influenced discovery of GPUs. If you'd prefer, leave that out and we can add it back in if it deemed to be necessary in the future.

garlick: > Actually, now that I think about this more, I'm not sure the binding options influenced discovery of GPUs. If you'd prefer, leave that out and we can add it back in if it deemed to be necessary in the future.

Should we contrive another question about cpu affinity that would let us keep those options documented in the FAQ and possibly show how the Flux cpu affinity options work and how to verify what's really going on with your app?  I'm not sure I have a great question in mind, so just throwing that out there.

For this PR I'll go ahead and leave that stuff out.

grondo: Probably, since that seemed to be the intent of this faq entry before

grondo: I guess the point to get across is that when flux discovers resources via hwloc, it honors the current core and gpu bindings, so if resources are missing, affinity and binding from the parent RM should be checked

garlick: Gotcha.  Just pushed a new question based on your comments.  It's close to the original question so I squashed the two commits into one that says it is splitting the original question (and renamed this PR).  Want to have a look and see if this works?

grondo: Yeah, your edits were already much clearer, but the new entry also handles what I think was implicitly answered before. Thanks!

grondo:
Feel free to set MWP (the previous comments were optional anyway)

PR #161, opened 2022-08-26
faqs: add more tips for launching many jobs

garlick:
Problem: the faqs entry "How do I efficiently launch a large
number of jobs" does not mention redirecting stdio to files, or
excluding rank 0 from scheduling either of which may be appropriate
in some cases.

Add bullets for these suggestions.

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -92,8 +92,8 @@ from the parent resource manager should be checked.  In Slurm, try
 How do I efficiently launch a large number of jobs?
 ===================================================
 
-If you have many jobs to run here are some tips to help improve efficiency
-and throughput:
+If you have more than 10K fast-cycling jobs to run, here are some tips that
+may help improve efficiency and throughput:
 
 - Create a batch job or allocation to contain the jobs in a Flux sub-instance.
   This improves performance over submitting them directly to the Flux system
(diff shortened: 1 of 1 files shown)


faqs.rst
```diff
@@ -115,6 +115,13 @@ and throughput:
   `Flux python interface <https://flux-framework.readthedocs.io/projects/flux-core/en/latest/python/index.html>`_.  The
   `flux-mini <https://github.com/flux-framework/flux-core/blob/master/src/cmd/flux-mini.py>`_
   command itself is a python program that can be a useful reference.
+- If jobs produce much output on stdout or stderr, use the
+  :core:man1:`flux-mini` ``--output`` and ``--error`` options to redirect them
```
grondo, line 119: `--output` redirects both stdout and stderr, so it is probably sufficient to just include that option in this short faq answer. (O/w, users might think they have to supply both to redirect both)

grondo (approved):
We may want to preface these bullets by mentioning that the following are probably only issues when running many thousands or tens of thousands of jobs, or if jobs produce a lot of output.

Do we want to also add a note about redirecting the content store away from /tmp   if the space available is small or is a ram filesystem? (Actually I seem to remember this being somewhere else?)

garlick:
Good suggestions. Just force pushed some changes that attempted to incorporate your comments.

Note that that there is another faq entry about filling up tmp:

https://flux-framework.readthedocs.io/en/latest/faqs.html#how-do-i-prevent-flux-from-filling-up-tmp

grondo:
> Note that that there is another faq entry about filling up tmp:

Thanks, I thought I remembered that. I wonder if we should put an "also see" bullet in this faq entry? Not required and this PR is already approved if you want to set MWP.

garlick:
Thanks!  I had that and removed it since the list of suggestions was already pretty long.  Let's leave off for now and we can always revisit later.

PR #164, opened 2022-09-07
reorganize top level index

garlick:
Problem: the fact that Flux is the native resource manager on some LLNL clusters may bring new users to the flux-docs index page that have a little different perspective than the researcher/workflow/early adopter people we seem to be expecting.  As such the organization and titles could use some tweaking.

This renames some top level pages and reorganizes the index.


cmoussa1 (approved):
LGTM! Nice reorganizing and improvement.

garlick:
Thanks guys - I'll do that.

PR #165, opened 2022-09-08
adminguide: add reminder to update DB after flux-accounting version upgrade

cmoussa1:
This PR adds a reminder in the Flux Accounting section in the Administrator's Guide to update the database on a system after upgrading to a newer version of flux-accounting where a DB already exists, as tables and columns could be added or removed when moving to a newer version.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -567,6 +567,16 @@ See also: :core:man5:`flux-config-job-manager`.
 Automatic Accounting Database Updates
 =====================================
 
+If updating flux-accounting to a newer version on a system where a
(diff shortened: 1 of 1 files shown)


garlick (approved):
Can that command handle upgrading from any past flux-accounting release or only the most recent release?  Just wondering if there is a constraint here that should be documented.

(LGTM though)

cmoussa1:
Yes, the command should be able to upgrade a DB from any past flux-accounting release.

Thanks for reviewing so quickly! I'll set MWP shortly

PR #166, opened 2022-09-08
adminguide: add example for heterogeneous nodes 

grondo:
@woodard was trying to set up a heterogeneous system instance of Flux and noticed there are no examples of how to generate R in this case.

Add a note about nodes with differing numbers of cores and an example of generating R with multiple `flux R encode`


garlick (approved):
Thanks, LGTM

PR #168, opened 2022-09-23
accounting: add multi-factor priority plugin section

cmoussa1:
This is a small PR that adds a new section to the flux-accounting guide which gives details about the multi-factor priority plugin, the factors it uses to calculate job priorities, and the job limits it currently enforces. It references the various job states defined in [RFC 21](https://github.com/flux-framework/rfc/blob/master/spec_21.rst) when talking about the job limit enforcement.

### DIFF
--- a/flux-accounting-guide.rst
+++ b/flux-accounting-guide.rst
@@ -138,6 +138,41 @@ This current job usage is then added to the association's previous job usage
 stored in the flux-accounting database. This sum then represents the
 association's overall job usage.
 
+****************************
+Multi-Factor Priority Plugin
+****************************
+
+The multi-factor priority plugin is a jobtap_ plugin that generates
+an integer job priority for incoming jobs in a Flux system instance. It uses
+a number of factors to calculate a priority and, in the future, can add more
+factors. Each factor has an associated integer weight that determines its
+importance in the overall priority calculation. The current factors present in
+the multi-factor priority plugin are:
+
+* **fair-share**: the ratio between the amount of resources allocated vs. resources
+  consumed.
+
+* **urgency**: a user-controlled factor to prioritize their own jobs.
+
+In addition to generating an integer priority for submitted jobs in a Flux
+system instance, the multi-factor priority plugin also enforces per-association
+job limits to regulate use of the system. The two per-association limits
+enforced by this plugin are:
+
+* **max_active_jobs**: a limit on how many *active* jobs an association can have at
+  any given time. Jobs submitted after this limit has been hit will be rejected
+  with a message saying that the association has hit their active jobs limit.
+
+* **max_running_jobs**: a limit on how many *running* jobs an association can have
+  at any given time. Jobs submitted after this limit has been hit will be held
+  by adding a ``max-running-jobs-user-limit`` dependency until one of the
+  association's currently running jobs finishes running.
+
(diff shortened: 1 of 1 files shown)


flux-accounting-guide.rst
```diff
@@ -138,6 +138,40 @@ This current job usage is then added to the association's previous job usage
...
+job limits to regulate use of the system. The two per-association limits
+enforced by this plugin are:
+
+* **max_active_jobs**: a limit on how many _active_ jobs an association can have at
+  any given time. Jobs submitted after this limit has been hit will be rejected
+  with a message saying that the association has hit their active jobs limit.
+
+* **max_running_jobs**: a limit on how many _running_ jobs an association can have
+  at any given time. Jobs submitted after this limit has been hit will be held
+  until one of the association's currently running jobs finishes running.
```
grondo, line 168: Perhaps note here that jobs "will be held by adding a `max-running-jobs` dependency" or similar.

flux-accounting-guide.rst
```diff
@@ -162,3 +196,7 @@ bank
 .. _Slurm accounting: https://slurm.schedmd.com/accounting.html
 
 .. _multi-factor priority plugin: https://slurm.schedmd.com/priority_multifactor.html
+
+.. _jobtap: https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html?highlight=jobtap
```
grondo, line 200: This link highlights every instance of the word "jobtap" in the docs, was that intentional? If not, maybe you wanted just:

https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html#flux-jobtap-plugins-7

grondo (commented):
Great addition! This is still a WIP so I'll just add a couple comments I noticed on a quick read.

flux-accounting-guide.rst
cmoussa1: Oh, that's a good suggestion. I'll go ahead and add this to the description, Thanks!

flux-accounting-guide.rst
cmoussa1: Oh I didn't even realize this! Thanks for pointing it out. I think there is only one instance of the word "jobtap" in the flux-accounting guide (with this PR), but I did not mean for it to highlight every potential future instance of the word. I'll change the link!

cmoussa1:
I just squashed and force-pushed a fix based on @grondo's suggestions above, and I'll take it out of [WIP] and mark it as ready for review.

flux-accounting-guide.rst
grondo: Oh, I meant that it highlights every instance of the word `jobtap` in the target of the link. And there's a lot of instance of `jobtap` in `flux-jobtap-plugins(7)` 😆

cmoussa1: Ohhhh, oops. Lol. Sorry for misunderstanding! Yeah, in that case, I definitely did _not_ want to have the original link that I used, right? So the link that you provided is better.

garlick (approved):
Looks great!  Just the one note on formatting I just mentioned in the coffee call.  Use *foo* to italicize, not _foo_.  All good otherwise, free free to fix that and set MWP.

cmoussa1:
Thanks for reviewing @garlick! I'll go ahead and set MWP on this then.

PR #169, opened 2022-10-04
admin guide: add flux-accounting DB permissions note

cmoussa1:
This PR adds a small note to the flux-accounting section in the Administator's Guide about correctly setting permissions and ownership for both the flux-accounting DB and the directory it resides in after it is installed on a system. It includes a reference to the SQLite documentation in this note about read/write permissions.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -543,21 +543,26 @@ parameters are assumed, including the accounting database path of
 
 .. code-block:: console
 
- $ flux account create-db
+ $ sudo -u flux flux account create-db
+
+.. note::
+    The flux accounting commands should always be run as the flux user. If they
+    are run as root, some commands that rewrite the database could change the
+    owner to root, causing flux-accounting scripts run from flux cron to fail.
 
 Banks must be added to the system, for example:
 
 .. code-block:: console
 
- $ flux account add-bank root 1
- $ flux account add-bank --parent-bank=root sub_bank_A 1
+ $ sudo -u flux flux account add-bank root 1
+ $ sudo -u flux flux account add-bank --parent-bank=root sub_bank_A 1
 
 Users that are permitted to run on the system must be assigned banks,
 for example:
 
 .. code-block:: console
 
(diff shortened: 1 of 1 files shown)


adminguide.rst
```diff
@@ -545,6 +545,13 @@ parameters are assumed, including the accounting database path of
 
  $ flux account create-db
 
+.. note::
+    In order to interact with the information located in the flux-accounting
+    database, permissions must be correctly set both for the database and the
+    directory it resides in, per the `SQLite documentation`_. Make sure to
+    change the ownership and read/write permissions to the appropriate group
+    and users before using any flux-accounting commands.
+
```
garlick, line 554: Since the instructions say to use flux cron to run the accounting scripts, and it runs as user `flux`, and those scripts need read-write access to the database, I _think_ the desired end result should be for the database to be owned by the user `flux` and at least mode 0600.  The directory it's in `/var/lib/flux` is created by systemd with owner `flux` and mode 0700, so it should  already let the `flux` user (re-)create the database.

It should be sufficient to change the instructions to prefix all the flux accounting commands we are asking poeple to run by hand with  `sudo -u flux`.

IMHO  it's still worthwhile to have a note box, but it should say something like 
> The flux accounting commands should always be run as the flux user.  If they are run as root, some commands that rewrite the database could change the owner to root, causing flux-accounting scripts run from flux cron to fail.

Note that we've done something with the `lcstaff` group on fluke - that needs an audit to make sure it's not going to blow up when a schema update occurs.

garlick (commented):
I

adminguide.rst
cmoussa1: Ah I see, sorry for misunderstanding you! Thanks for explaining the reasoning too.

I've just squashed and force-pushed a fix to add `sudo -u flux` to all of the flux-accounting commands as well as changed the note to include your suggestion.

garlick (approved):
Thanks @cmoussa1!  LGTM.

cmoussa1:
Thanks! I'll go ahead and set MWP on this

PR #170, opened 2022-10-05
adminguide: update for flux-core 0.44.0, add background

garlick:
This updates the caveats for the just-released flux-core 0.44.0, adds a subsection describing how to set up queues, and adds some background to the overview section.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -20,9 +20,12 @@ resource manager on a cluster.
     a 128 node cluster.
 
 
-********
-Overview
-********
+***********************
+Overview and Background
+***********************
+
+Flux Architecture
+=================
 
 The base component of Flux is the :core:man1:`flux-broker` executable.  A Flux
 instance consists of one or more ranked Flux brokers communicating over a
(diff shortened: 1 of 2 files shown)


adminguide.rst
```diff
@@ -74,6 +77,92 @@ with operation differentiated only by configuration.
...
+  Each broker module runs in its own thread as part of the broker executable,
+  communicating with other components using messages.  Broker modules are
+  dynamically loadable with the :core:man1:`flux-module` command.  Core
+  services like the KVS, job manager, and scheduler are implemented using
+  broker modules.
+
+jobtap plugins
+  The job manager orchestrates a job's life cycle.  Jobtap plugins extend the
+  job manager, arranging for callbacks at different points in the job life
+  cycle, similar to Slurm's SPANK plugins.  Jobtap plugins may be dynamically
```
grondo, line 98: I would remove the comparison to SPANK (since SPANK plugins are more like shell plugins, not jobtap plugins, which would be plugins within the slurmctld)

adminguide.rst
```diff
@@ -200,9 +200,8 @@ flux-core
   prioritization. If building flux-core from source, be sure to configure with
   ``--with-flux-security``. Install on all nodes (required).
 
-flux-sched (optional)
-  The Fluxion graph-based scheduler.  Install on management node
-  (optional, but recommended for production multi-user system installs).
+flux-sched
```
grondo, line 203: I don't have any issues with this change, but for my own curiosity - the commit message says "fluxion ... is required for queues". Is that true? I guess you state that because queues used to partition cannot work with a FCFS scheduler, or are there other features missing to implement queues (at least for testing) in flux-core alone?

grondo (approved):
LGTM! (Sorry I accidentally posted an individual comment not part of the review)

adminguide.rst
garlick: Oops sorry for that misreference.  I'll remove.

adminguide.rst
garlick: That might have been too strong of a statement.  Yes, that problem.  I'll reword.

garlick:
Setting MWP.

PR #172, opened 2022-10-08
move accounting section of admin guide into accounting guide

garlick:
Tell me what you think about this @cmoussa1.  I thought it might be better to have all the accounting information in one manual rather than splitting it between the admin guide and the accounting guide, so this consolidates it in the accounting guide. This also adds an intro section which seemed necessary to establish some context.

(Please read the intro carefully and make sure I'm not misrepresenting the project or misusing the terminology!)

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -268,8 +268,9 @@ flux-sched
   The Fluxion graph-based scheduler.
 
 flux-accounting (optional)
-  Accounting database of user/bank usage information, and a priority plugin.
-  Install on management node (early preview users only).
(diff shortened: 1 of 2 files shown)


cmoussa1 (approved):
Thanks for putting this together @garlick! I think this is a great idea to have the flux-accounting information consolidated into one guide. I gave the intro a couple of read-throughs and thought it did a good job introducing the project, so this LGTM!

garlick:
Thanks!  OK, I'll set MWP and feel free to change later on as needed.

PR #177, opened 2022-11-02
adminguide: minor tune-up of overview architecture

garlick:
Problem: the Overview architecture sub-section opens with a sentence about the Flux broker executable, when what is trying to convey is the notion of a Flux instance as a collection of brokers.  It seems like an odd way to begin.

Reword that sentence to lead with the "flux instance" term. Also reword another sentence in that section that seemed awkward.

### DIFF
--- a/adminguide.rst
+++ b/adminguide.rst
@@ -27,11 +27,10 @@ Overview and Background
 Flux Architecture
 =================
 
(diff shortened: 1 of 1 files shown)


grondo (approved):
Nice improvement!

PR #178, opened 2022-11-03
adminguide: add flux-pam

garlick:
Problem: the admin guide does not mention the flux-pam package.

Add flux-pam to the package list, and add a brief section on configurign it (mostly just a placeholder).

Rename the exsiting "Configuring PAM" section to be more specific to the IMP PAM config.


grondo (approved):
LGTM! Thanks.

Issue #186, opened 2023-01-17
docs: add glossary, terms section


cmoussa1:
Mentioned in #185, it might be a good idea to add a **Glossary** page to our docs site to go into slightly more detail with some of the Flux-specific concepts we introduce in our guides. This could either be a separate page, hyperlinks to footnotes on the bottom of the page they are introduced in, or both.

Definitions of these terms should probably be kept brief; if they are complex terms or concepts, perhaps there should be consideration of making them their own page on the docs site.

I think it might be a good idea to collaborate some on coming up with these definitions (since I am not the one who came up with a number of these concepts and am probably bound to get them wrong 😆), so if others are okay with it, this could make for a good coffee hour exercise! I would be happy to work on adding these to the docs site, however.

garlick:
A few quick ideas
- flux instance
- system instance
- single user mode
- job eventlog
- kvs
- resource definition (R)
- jobspec
- workflow
- high throughput computing
- fractal scheduling (scheduler nesting)
- PMI
- flux-imp
- jobtap plugin
- flux-broker
- flux-shell
- job request frobnicator
- job request validaton
- feasability
- hostlist
- idset
- flux standard duration
- FLUID
- TOML
- perilog
- queue terms (start/stop/enable/disable)
- policy
- limits
- drain
- overlay terms (full, partial, degraded, lost, offline)

chu11:
i'll just throw this out there, would a similar "acronyms" page be useful? or is a glossary enough.

tbon
kvs
imp
fluid
R

ehhh the list is sorta short, perhaps longer out if we end up with enough of them

cmoussa1:
I think that's a great idea. If we don't dedicate an entire page, we could at least spell out the acronym in parentheses the first time it is used in our docs. This is at least the format I follow when I write papers.

For example, the _Flux KVS_ section already does something like this:

> The key-value store (kvs)

Although the method I mentioned above reverses this:

```diff
- The key-value store (kvs)
+ The KVS (key-value store)
```

garlick:
IMHO throw everything in the glossary, acronym or backronym or no.

PR #185, opened 2023-01-14
first shot at re-organizing sections

vsoch:
This is a high level re-org of the docs structure! Specifically I've:

 - made separate directories to organize jobs, guides, and tutorials
 - added a few more FAQ to introduce to flux (and a few more categories)
 - simplified a few page headers so they are less texty (we likely will want to do more of this)

I think having these sections will make it easier to, for example, know where to add an LLNL-cluster-specific tutorial, or a link to an integration tutorial, etc. Here are some screenshots of the new TOC and various sections (and some critique for future work):

image

My critique here is that (although it makes sense to start with install or build) we jump into the quick start assuming the user can relate to these sections. E.g., what the heck is a flux instance? And KVS? What is needed here is an opening paragraph that says "here we are going to install, create X (which is defined as) and then do this (another definition). By the time they get to the sections like "Flux KVS" it should be familiar. We might also rename these headers to be general. E.g., instead of Flux KVS we would say something like "Flux Cluster Information".

Here is the new FAQ - I've added more sections

image

And tutorials, now also organized

image

And jobs

image

And guides!

image

This should be a high level set of changes, so for discussion here let's talk about the different kinds of content we would eventually want, and if this organization strategy makes sense for that.

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

PR #189, opened 2023-01-19
docs: add a new "Glossary" section

cmoussa1:
#### Background

The documentation site contains terms, both Flux-specific and not, that might be unclear to users who are reading the site and/or about Flux for the first time. It was discussed in #185 that the site might benefit from having a page dedicated to explaining in slightly more detail some of the terms, concepts, and acronyms mentioned throughout our site which may be unfamiliar to users who are reading about Flux (or resource managers, HPC concepts, etc.) for the first time.

---

This is a small PR meant to kick off the start of a **Glossary** page for our docs site, a page that explains in more detail these terms and concepts. Mention of this new page is referenced at the top of the **Quick Start** guide for users who might want to look back at this page as they read throughout the guide and other pages on our sites.

A few definitions are also added to this page, but they surely do not make up _all_ of the terms we might want to eventually include. Furthermore, if folks who have feedback/improvements to the terms I've defined in this PR, please let me know. I myself am not familiar with all of the terms suggested in #186 and am pretty sure I did not do all of the definitions their proper justice. 😅 I am simply just trying to start us off!

### DIFF
--- a/glossary.rst
+++ b/glossary.rst
@@ -1,0 +1,125 @@
+.. _glossary:
+
+========
+Glossary
+========
+
+This page contains definitions for Flux-specific terms and concepts mentioned
+throughout the various guides and docs on our sites.
+
+------------------------------------------------------------------------------
+
+.. _flux-broker:
+
+**Flux Broker**
+  A distributed message broker that provides communications services
+  within a Flux instance. It is the main controller of the Flux instance and
+  loads all the important job, scheduling, and resource related services for
+  Flux.
+
+.. _flux-instance:
+
+**Flux Instance**
+  A group of flux-broker processes which are launched via any parallel launch
+  utility that supports PMI (Process Manager Interface).
+
+.. _eventlog:
+
+**Eventlog**
+  An ordered log of timestamped events, stored in the Flux KVS and defined by
+  `RFC 18 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_18.html>`_.
+  Eventlogs are used to record job events, capture standard I/O streams,
+  and record resource status changes.
+
+.. _fluid:
+
+**FLUID (Flux Locally Unique ID)**
+  IDs used to represent Flux jobs, defined by
+  `RFC 19 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_19.html>`_.
+
+.. _hostlist:
+
+**Hostlist**
+  A format supported by many HPC tools to create a compact list of hosts by
+  name, consisting of a common prefix, a number, and an optional suffix. For
+  example, the hosts ``cow1``, ``cow2``, and ``cow3`` could be represented by
+  the hostlist ``cow[1-3]``. 🐮
+
+.. _idset:
+
+**IDSet**
+  A compact way of representing an unordered set of integers, defined by
+  `RFC 22 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_22.html>`_.
+
+.. _jobspec:
+
+**Jobspec**
+  A file which defines the resources, tasks, and other attributes of a single
+  program, defined by `RFC 25 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_25.html>`_
+  and `RFC 14 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_14.html>`_.
+  An example can be found in the `Example Jobspec
+  <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_25.html#example-jobspec>`_
+  section of RFC 25.
+
+.. _jobtap-plugin:
+
+**Jobtap Plugin**
+  A built-in or external plugin that can be loaded into the job manager broker
+  module. They are an extension of the job manager, arranging for callbacks at
+  different points in a job's life cycle. Additional information on jobtap plugins
+  can be found in :core:man7:`flux-jobtap-plugins`.
+
+.. _kvs:
+
+**KVS (Key-Value Store)**
+  A component of a Flux instance where information pertaining to the instance
+  is stored in a key-value format. It is used by Flux internally to store data
+  such as resource information, jobspec, logs, and checkpoints. Each job is
+  assigned a KVS namespace, which may also be used by users to store small
+  amounts of data.
+
+.. _limits:
+
+**Limits**
+  Maximum values that are enforced across jobs, queues, and users, such as time
+  and/or resource limits.
+
+.. _pmi:
+
+**PMI (Process Manager Interface)**
+  A standard API and wire protocol for communication between MPI runtimes and
+  resource managers.
+
+  PMI traditionally provides a key-value store with all-to-all style synchronization
+  to support the bootstrap of parallel programs such as MPI. During wire up, each
+  parallel task can bind to a network endpoint, put its network address into the PMI
+  KVS, synchronize, and then get the addresses of its network peers. All without
+  requiring the process manager (e.g. Flux) to know anything about the network
+  interconnect hardware. Brilliant!
+
+  Although there are several variations of PMI, Flux implements the original one
+  from Argonne National Lab which is documented in
+  `RFC 13 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_13.html>`_.
+  In addition to providing PMI service to the programs it launches, Flux also uses
+  PMI for its own wire-up when it is launched by Flux or a foreign launcher.
+
+  PMI was an unfortunate three letter acronym choice, given that MPI was already a
+  thing. This is known - no whining.
+
+.. _policy:
+
+**Policy**
+  Attributes of a Flux queue that define behavior of the queue, such as time or
+  resource limits.
+
+.. _r:
+
+**R**
+  A resource set defined in JSON (`Javascript Object Notation <https://json-spec.readthedocs.io/reference.html>`_)
+  format, defined by `RFC 20 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_20.html>`_.
+
+.. _shell-plugin:
+
+**Shell Plugins**
+  Extensions of a job's environment that can be configured on a per-job basis
+  using the ``--setopt`` option of the :core:man1:`flux-mini` commands.
--- a/quickstart.rst
+++ b/quickstart.rst
@@ -4,7 +4,9 @@
 Quick Start
 ============
 
-A quick introduction to Flux and flux-core.
+A quick introduction to Flux and flux-core. For reference, the :ref:`Glossary`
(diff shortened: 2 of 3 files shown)


glossary.rst
```diff
@@ -0,0 +1,46 @@
...
+========
+Glossary
+========
+
+This page contains definitions for Flux-specific terms and concepts mentioned
+throughout the various guides and docs on our sites.
+
+------------------------------------------------------------------------------
+
+**Flux Broker**
```
vsoch, line 12: Should we have references to each of the terms so they can be linked (specifically) from other pages?

glossary.rst
```diff
@@ -0,0 +1,46 @@
...
+  A distributed message broker daemon that provides communications services
```
vsoch, line 13: I would say something in layman's terms like:

> You can imagine the broker as the main controller of the Flux instance - the one running the show, and ensuring that X, Y, Z.

(Metaphors are generally great for docs, when you take a concept and turn it into a mini story so it sticks better).

glossary.rst
```diff
@@ -0,0 +1,46 @@
...
+  within a Flux instance.
+
+**Flux Instance**
+  A group of flux-broker processes which are launched via any parallel launch
+  utility that supports PMI (Process Manager Interface).
+
+**Hostlist**
```
vsoch, line 20: Can we show an example here? I would say it's a list of hosts with patterns that are known how to be expanded (and show an example).

glossary.rst
```diff
@@ -0,0 +1,46 @@
...
+  A format supported by existing HPC tools to create a compact list of hosts by
+  name, consisting of a common prefix, a number, and an optional suffix.
+
+**Jobspec**
```
vsoch, line 24: Definitely need an example here, even if simple! And then to say what the Jobspec specifically does. E.g., what is creating it? If it's a program, is it being launched somewhere?

glossary.rst
```diff
@@ -0,0 +1,46 @@
...
+  A file which defines the resources, tasks, and other attributes of a single
+  program.
+
+**Jobtap Plugin**
```
vsoch, line 28: An example?

glossary.rst
```diff
@@ -0,0 +1,46 @@
...
+  A built-in or external plugin that can be loaded into the job manager broker
+  module.
+
+**KVS**
```
vsoch, line 32: Probably should start with "Key Value Store" to say what it means. And add what people use it for (users? developers? both?)

glossary.rst
```diff
@@ -0,0 +1,46 @@
...
+  A component of a Flux instance where information pertaining to the instance,
+  such as resource information, is stored in a key-value format.
+
+**Limits**
+  Maximum values that are enforced across jobs, queues, and users, such as time
+  and/or resource limits.
+
+**PMI**
```
vsoch, line 40: What does PMI mean? Why do we use it?

quickstart.rst
```diff
@@ -4,7 +4,9 @@
 Quick Start
 ============
 
-A quick introduction to Flux and flux-core.
+A quick introduction to Flux and flux-core. For reference, the :ref:`Glossary`
```
vsoch, line 7: I think this is good for the first shot - what we will eventually do is define a term later and then put a link to it so the person can click and immediately be taken to the exact definition. We can also look into if it might be possible to do some kind of popup so they don't leave the page.

vsoch (commented):
Nice!! I left some comments, and let's get others' feedback too.

glossary.rst
cmoussa1: I like this suggestion a lot, thanks! It might be helpful for me to get an example of what _X,Y,Z_ might be in this metaphor (maybe the flux-core guys can help me out here).

glossary.rst
cmoussa1: Sure! That sounds like a good suggestion to me. Are you thinking of an example at the end of the definition like the following? "Example: `my_host[1-10]`"

glossary.rst
vsoch: haha, I put letters because I'm not totally sure. I could guess based on the flux operator, but best to ask. If it's helpful here is the template I use:

```yaml
# Flux needs to know the path to the IMP executable
[exec]
imp = "/usr/libexec/flux/flux-imp"

[access]
allow-guest-user = true
allow-root-owner = true

# Point to resource definition generated with flux-R(1).
[resource]
path = "/etc/flux/system/R"

[bootstrap]
curve_cert = "/etc/curve/curve.cert"
default_port = 8050
default_bind = "tcp://eth0:%%p"
default_connect = "tcp://%%h.%s:%%p"
hosts = [
	{ host="%s-%s"},
]
```

glossary.rst
vsoch: Yes exactly! Say something like `my_host[1-10]` expands to `my_host1` `my_host2` ... `my_host10`

glossary.rst
cmoussa1: Hmmm, I wonder if there's an example small enough to paste on this page. I guess I could paste a link to the [flux-accounting jobtap plugin](https://github.com/cmoussa1/flux-accounting/blob/master/src/plugins/mf_priority.cpp) or the examples in the [flux-core repo](https://github.com/flux-framework/flux-core/tree/master/t/job-manager/plugins) (thinking about it out loud, linking the latter might make more sense as those are written in C, like the rest of flux-core). What do you think?

glossary.rst
cmoussa1: Oh, yes, can't believe I forgot to spell out the acronym (when that's what we talked about why the Glossary page was necessary 🤦).

> And add what people use it for (users? developers? both?)

Good point. Will also ping this to the flux-core guys to get more explanation.

glossary.rst
cmoussa1: I should spell this out too (PMI stands for (I believe) **Process Manager Interface**). This definition is also something I think I could use some help on from the flux-core guys ^

glossary.rst
vsoch: I think that would work for now - in the future for developer docs we will want documentation that unwraps what it means to be a plugin, how to write one, etc.

glossary.rst
garlick: Maybe add:
> PMI traditionally provides a key-value store with all-to-all style synchronization to support the bootstrap parallel programs such as MPI.   During wire up, each parallel task can bind to a network endpoint, put its network address into the PMI KVS, synchronize, and then get the addresses of its network peers.  All without requiring the process manager (e.g. Flux) to know anything about the network interconnect hardware.  Brilliant!
>
> Although there are several variations of PMI, Flux implements the original one from Argonne National Lab which is  documented in [RFC 13](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_13.html).   In addition to providing PMI service to the programs it launches, Flux also _uses_ PMI for its own wire-up when it is launched by Flux or a foreign launcher.
>
> PMI was an unfortunate three letter acronym choice, given that MPI was already a thing.  This is known.

chu11:
I think this is an awesome start.  Once we get this merged, we can all start adding a few more terms here and there.  I'm already gearing for some of my additions!

glossary.rst
garlick: Maybe some of the terms and definitions listed here could be just slotted in as a first pass?

https://flux-framework.readthedocs.io/en/latest/guides/admin-guide.html#software-components

glossary.rst
grondo: One or more of the [test vectors](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_29.html#test-vectors) from the RFC might work?

glossary.rst
grondo: there are some simple jobtap plugins used for testing in the flux-core testsuite. However, I'm not convinced copy and paste of some C code would elucidate much here since even a simple plugin might be quite verbose. Would it be enough to paste a link to the `flux-jobtap-plugins(7)` man page?

cmoussa1:
OK, I just force-pushed some changes that (hopefully) should address most of the great feedback and suggestions I received above. These changes include the following:

* references are now added to every entry on the Glossary page so that they can be linked to throughout the site
* some of the definitions have been reworked to hopefully be more explanatory (thanks for the suggestions on definitions above and in Slack DMs 😄)
* the terms in the Glossary that are acronyms have their abbreviations spelled out when they are mentioned for the first time
* hyperlinks are added to the first mention of some of the Flux terms in the Quickstart page that redirect the user to the definition of the term on the Glossary page

glossary.rst
```diff
@@ -0,0 +1,125 @@
...
+  IDs used to represent Flux jobs, defined by
+  `RFC 19 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_19.html>`_.
+
+.. _hostlist:
+
+**Hostlist**
+  A format supported by existing HPC tools to create a compact list of hosts by
+  name, consisting of a common prefix, a number, and an optional suffix. For
+  example, the hostlist ``"foo[1-5]"`` can be expanded to
+  ``"foo1,foo2,foo3,foo4,foo5"``.
```
chu11, line 46: instead of "existing", "many"?

chu11: Perhaps say "the hosts foo1,foo2, and foo3 could be represented by the hostlist foo[1-3]."?

vsoch: Yeah I like that! And maybe go nuts and pick a fun word like fry, cow, cat, bun, fun.... there are lots of good ones! https://www.playosmo.com/kids-learning/3-letter-words-for-kids/ :laughing: 

:potato: :cow: :cat: :bread: :tada: 

Extra points for a matching emoji!  A nicely placed emoji, maybe one per page of docs, makes them more fun.

cmoussa1: Cool! Thanks to you both for the suggestions. I just force-pushed a change that re-worded the example at the end of the definition for hostlist. I chose `cow` for our pretend cluster, lol. 🐮

glossary.rst
```diff
@@ -0,0 +1,125 @@
...
+  PMI for its own wire-up when it is launched by Flux or a foreign launcher.
+
+  PMI was an unfortunate three letter acronym choice, given that MPI was already a
+  thing. This is known - no whining.
+
+.. _policy:
+
+**Policy**
+  Attributes of a Fux queue that define behavior of the queue, such as time or
+  resource limits.
```
chu11, line 113: Fux -> Flux

chu11 (approved):
LGTM, just noticed the one typo below.

glossary.rst
cmoussa1: Thanks, just force-pushed a change to fix this typo.

cmoussa1:
Thanks for approving @chu11! @vsoch, did you want to give this a PR another look?

vsoch (approved):

cmoussa1:
Woo! Thanks. Setting MWP 🎉

PR #190, opened 2023-01-27
Add manager comparison table

vsoch:
This is a WIP! I have a css/javascript bug somewhere that the last few rows are missing colors, but wanted to share the WIP for quick feedback, etc.

image

Unless I embedded some html table library to customize this, I couldn't find a way for the simple list-table in rst to support having full width rows every once in a while, so I opted for three separate tables (and I think it looks OK?)  I'm also planning on adding some text that guides the reader to the right places if they want to suggest a feature to add, or point out we are wrong about something.

Finally, note that the same code renders on its own page (above) and the FAQ (below)

image

Here is the original png for comparison.

image

I'm going for my run because it's been a day, but I'll be back and hopefully have more patience for JavaScript after that. :laughing:  Back in a bit!

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -39,6 +39,12 @@ you can easily demo Flux in a container, or even used in Kubernetes with the
 `Flux Operator <https://flux-framework.org/flux-operator>`_. We have a vision for Flux to
 allow for converged computing, or making it easy to move between traditional HPC and cloud.
 
+How does Flux compare to other managers?
+========================================
+
+Flux has similarities to many other resource managers, but we do a variety of things quite differently. 
+Check out this comparison table here (:ref:`comparison-table`) for a detailed list!"
+Do you want to add or correct a feature? Please `let us know <https://github.com/flux-framework/flux-docs/issues>`_
 
 *****************
 General Questions
--- a/includes/comparison-table.html
+++ b/includes/comparison-table.html
@@ -1,0 +1,51 @@
+<style>
+.comparison-table {
+  width: 100% !important;
+}
+</style>
+</style>
+<script src="https://code.jquery.com/jquery-3.6.3.slim.min.js" integrity="sha256-ZwqZIVdD3iXNyGHbSYdsmWP//UBokj2FHAxKuSBKDSo=" crossorigin="anonymous"></script>
+<script>
+$(function() {    
+  var tables = document.getElementsByClassName('comparison-table') 
+  for (var t=0, len=tables.length; t<len; t++){
+    var table = tables[t]
+      if (!table){
+        continue
+      }
+      var cells = table.getElementsByTagName('td');
+      for (var i=0, len=cells.length; i<len; i++){
+        cell = cells[i]
+        if (!cell){
+          continue
+        }
+        var text = cells[i].outerText
+        console.log(text)
+        if (text.includes("no")){
+          cells[i].style.backgroundColor = '#ff8c8c';
+        }
+        if (text.includes("yes")){
+          cells[i].style.backgroundColor = "#96f096";
+        }
+        if (text.includes("outdated")){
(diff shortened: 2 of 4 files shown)


vsoch:
okay fixed the bugs, this is ready for review! This is maybe the one thing I didn't epically fail on today. :sob: :laughing: 

image
image

vsoch:
Preview! https://flux-framework--190.org.readthedocs.build/en/190/tables/comparison-table.html and the table is also in the FAQ. I think questions for discussion are:

- where on the site we should put the table vs link it (e.g., I'm not sure it belongs fully in the FAQ but the link makes sense
- should it also be part of the portal (or another FAQ that links to it from there?)

cmoussa1:
Wow this looks really awesome! Great work on this so far @vsoch. 👍

> where on the site we should put the table vs link it (e.g., I'm not sure it belongs fully in the FAQ but the link makes sense

IMHO, something substantial like this probably warrants its own page on the site instead of just a FAQ entry. However, I definitely think it's worth mentioning a table like this on the FAQ page and providing a link to it. So, under the question "How does Flux compare to other resource managers?" an answer could say something like:

"Flux has similarities to many other resource managers, but we do a variety of things quite differently. Check out this comparison table **here** (insert hyperlink) for a detailed list!"

Or maybe something along those lines. I've never been a strong writer 😅 lol

> should it also be part of the portal (or another FAQ that links to it from there?)

Just my thoughts, but keeping a table like this in just one spot makes the most sense to me. Especially if we want to add/edit rows or columns in the future. We could always link to it throughout the portal wherever we feel appropriate.

vsoch:
> IMHO, something substantial like this probably warrants its own page on the site instead of just a FAQ entry. 

Yep - it has it's own page! I will update the table on the FAQ page to just be the link, and update with the text you provided.

Changes coming soon!

vsoch:
Done!

vsoch:
I'm so excited to merge! I can't wait to share with everyone. I spent like half a day on this table, specifically figuring out how to make it, lol.

_insert Jurassic Park reference... she's learning!!_

garlick:
That really did turn out nice!  Thanks @vsoch !

grondo:
One quick question about the content. The table row labeled `Bulk job submission` states "Uniform jobs only" for Slurm, but though I could be confused about what is meant by bulk job submission in this context, Slurm does have [Job Arrays](https://slurm.schedmd.com/job_array.html), which seem to be a non-uniform bulk submission feature.

Not sure if that warrants a change in that row for Slurm since I'm not sure of the intent there, but thought I'd point it out since the table should at least appear fair...

garlick:
The job array description says
> All jobs must have the same initial options (e.g. size, time limit, etc.), however it is possible to change some of these options after the job has begun execution using the scontrol command specifying the JobID of the array or individual ArrayJobID.

So I can see where the authors might think the statement is fair.  (I'm not knowledgeable about this feature of slurm so I guess I'm not adding much here)

chu11:
dumb nuance question, if there is only one language binding (Moab's "some (C)"), shouldn't the column just say "no"?  Perhaps my interpretation of "binding" is different than others.  But the way I think of it, there's the "base" language and bindings against that language.  So if only the base language is supported, there are no bindings?

vsoch:
I think for this PR it should be scoped to adding the current (published?) table, and any changes that are wanted should be followed up separately.

garlick:
Agreed.  Feel free to set the merge when passing tag for mergify since @cmoussa1 already approved.

vsoch:
Can I merge it then? :D

vsoch:
oh it already happened, yay!!!

garlick:
Mergify does that for you and rebases on current master if needed.

PR #191, opened 2023-02-01
Add more glossary terms

chu11:
Some were listed in #186, some weren't.  Just trying to fill it out more.

### DIFF
--- a/glossary.rst
+++ b/glossary.rst
@@ -9,6 +9,14 @@ throughout the various guides and docs on our sites.
 
 ------------------------------------------------------------------------------
 
+.. _eventlog:
+
+**Eventlog**
+  An ordered log of timestamped events, stored in the Flux KVS and defined by
+  `RFC 18 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_18.html>`_.
+  Eventlogs are used to record job events, capture standard I/O streams,
+  and record resource status changes.
+
 .. _flux-broker:
 
 **Flux Broker**
@@ -22,14 +30,6 @@ throughout the various guides and docs on our sites.
 **Flux Instance**
   A group of flux-broker processes which are launched via any parallel launch
   utility that supports PMI (Process Manager Interface).
-
-.. _eventlog:
-
-**Eventlog**
-  An ordered log of timestamped events, stored in the Flux KVS and defined by
-  `RFC 18 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_18.html>`_.
-  Eventlogs are used to record job events, capture standard I/O streams,
-  and record resource status changes.
 
 .. _fluid:
 
--- a/glossary.rst
+++ b/glossary.rst
@@ -25,17 +25,50 @@ throughout the various guides and docs on our sites.
   loads all the important job, scheduling, and resource related services for
   Flux.
 
+.. _flux-imp:
+
+**Flux IMP (Independent Minister of Privilege)**
+  A system service that allows instance owners to run work on behalf of a guest.
+  In other words, it allows multiple users to use / run jobs on a :ref:`Flux
+  instance<flux-instance>` instead of just the owner.  More of the nitty gritty
+  details can be found in
+  `RFC 15 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_15.html>`_.
+
 .. _flux-instance:
 
 **Flux Instance**
   A group of flux-broker processes which are launched via any parallel launch
   utility that supports PMI (Process Manager Interface).
 
+.. _flux-shell:
+
+**Flux Shell**
+  A component of Flux that manages the startup and execution of user jobs.  You
+  can read more in the :core:man1:`flux-shell` manpage.
+
 .. _fluid:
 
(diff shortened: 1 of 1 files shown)


glossary.rst
```diff
@@ -17,26 +25,51 @@ throughout the various guides and docs on our sites.
...
-.. _eventlog:
+.. _flux-shell:
 
-**Eventlog**
-  An ordered log of timestamped events, stored in the Flux KVS and defined by
-  `RFC 18 <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_18.html>`_.
-  Eventlogs are used to record job events, capture standard I/O streams,
-  and record resource status changes.
+**Flux Shell**
+  A component of Flux which manages the startup and execution of user jobs.  You
```
vsoch, line 46:
```suggestion
  A component of Flux that manages the startup and execution of user jobs.  You
```

glossary.rst
```diff
@@ -84,6 +117,41 @@ throughout the various guides and docs on our sites.
   Maximum values that are enforced across jobs, queues, and users, such as time
   and/or resource limits.
 
+.. _queue:
+.. _job_queue:
+
+**Queue**
+  A container for a set of jobs in priority order.  Queues may differ in what
```
vsoch, line 124: I think some might read this as "linux container" but I don't think that's what you mean. Is there another term we can use?

glossary.rst
```diff
@@ -84,6 +117,41 @@ throughout the various guides and docs on our sites.
...
+  resources are assigned to them, which users are allowed to submit to them, and
+  what defaults/limits may exist for them.  Often may be called a "job queue".
```
vsoch, line 126:
```suggestion
  what defaults/limits may exist for them.  Often may be called a "job queue."
```

glossary.rst
```diff
@@ -84,6 +117,41 @@ throughout the various guides and docs on our sites.
...
+
+.. _queue_enable:
+.. _queue_disable:
+.. _queue_start:
+.. _queue_stop:
+
+**Queue Enable/Disable/Start/Stop**
+  In Flux, the terms "enable"/"disable" are used to describe if jobs can be
+  submitted to a queue.  The terms "start"/"stop" are used to describe if jobs
+  can be executed.  Under normal operations, a queue is "enabled"/"started",
```
vsoch, line 136:
```suggestion
  can be executed.  Under normal operations, a queue is "enabled"/"started,"
```

glossary.rst
```diff
@@ -84,6 +117,41 @@ throughout the various guides and docs on our sites.
...
+  which allows users to submit jobs and jobs can run.  If a queue is
+  "enabled"/"stopped", it means jobs can be submitted but not allowed to run.
```
vsoch, line 138:
```suggestion
  "enabled"/"stopped," it means jobs can be submitted but not allowed to run.
```

glossary.rst
```diff
@@ -84,6 +117,41 @@ throughout the various guides and docs on our sites.
...
+  This siutation is common if resources are temporarily down for maintenance.
+  If a queue is "disabled"/"started", previously submitted jobs can run, but no
```
vsoch, line 140:
```suggestion
  If a queue is "disabled"/"started," previously submitted jobs can run, but no
```
https://www.thesaurus.com/e/grammar/does-punctuation-go-inside-or-outside-quotation-marks/

glossary.rst
```diff
@@ -123,3 +191,51 @@ throughout the various guides and docs on our sites.
...
+  `Wikipedia page <https://en.wikipedia.org/wiki/Uniform_Resource_Identifier>`_
+  for general overview.
+
+.. _workflow:
+
+**workflow**
+  The
+  `Webster's dictionary <https://www.merriam-webster.com/dictionary/workflow>`_
+  definition is "the sequence of steps involved in moving from the beginning to
+  the end of a working process".  In the context of Flux, this is usually a
```
vsoch, line 237:
```suggestion
  the end of a working process."  In the context of Flux, this is usually a
```

glossary.rst
chu11: Hmmm.  I used "container" b/c that's what we use in the RFC.   Doing nothing more than googling "what is a job queue", first description seems to use "data structure", which I don't think is quite right for this context.

Will ponder, perhaps could re-word the whole thing in a different way.

vsoch: Maybe something simple and stupid like a list? Or another synonym for queue?

chu11: none of the synonyms spoke to me (vessel, receptacle, ...).  As I went for a walk, "data structure" began sounding better to me.

"A scheduling data structure that contains sets of jobs in priority order."

chu11:
re-pushed with fixes / adjustments from comments above

glossary.rst
vsoch: Went for a walk, you say? I have a song for you! https://youtu.be/dZX6Q-Bj_xg

I think data structure sounds ok too, at least I won't confuse it with "those other container things."

chu11:
rebased and re-pushed, "sub-instance" -> "subinstance"

vsoch (approved):
I think this looks good.

PR #195, opened 2023-02-08
[WIP] fast_job_submission: add new guide

chu11:
Add new guide to improve job submission speed in Flux.

Note, for now I just stuck this stupidly under "jobs".  I'm pretty sure its not the right place.

Possibly a "guide" section under tutorials?  Ehhh?


chu11:
re-pushed, I decided to put the tutorial under a new "how to" section of tutorials.  The idea is that the "how to" topics will likely involve multiple commands or multiple prior topics, used together to describe something.

vsoch:
@chu11 I think (my preference) for the time being since we have an entire jobs section would be to keep the tutorial under there, and when it's warranted to re-organized we think of all the pages in this section in that context. We are also working on the Command tutorials section, and I think it's going to get too watered down to start with "Howto" and "Commands." Arguably, everything is a "How to" and it's a category that moreso reflects we don't know exactly where to put it (and I'm going to suggest the well-established Jobs section for now).

chu11:
yeah, i feel its a little bit in between the two.  But lets go with the jobs section for now, moved it back.

chu11:
rebased and re-pushed, "sub-instance" -> "subinstance"

vsoch (approved):
I think this tutorial is great and I don't see any issues, however I'm not experienced with this kind of submission.

chu11:
re-pushed, cut and pasted `code-block:: sh` in some places where it should be code-block:: console`.

chu11:
> I think this tutorial is great and I don't see any issues, however I'm not experienced with this kind of submission.

Perhaps @grondo or @garlick could take a quick skim, make sure what I'm writing sounds good.

chu11:
re-pushed, with comment fix above and removing "flux mini" references

PR #199, opened 2023-02-10
tutorials: correct code blocks colorization

chu11:
Problem: The code blocks where syntax highlighted to 'sh', but they should just be 'console' output.

Update all code blocks to be 'console' color outputted.

---

@vsoch - agree?  I just noticed some weird colorization in the console output.


vsoch (approved):
Agree - and this is my bad for copying the style of blocks from other parts of the flux-docs!

chu11:
FYI, normally the PR author sets `merge-when-passing`  just in case they want to adjust a commit message or something like that.

PR #200, opened 2023-02-10
tutorials: add flux proxy command tutorial

chu11:


tutorials/commands/index.rst
```diff
@@ -6,8 +6,9 @@ Command Tutorials
 Welcome to the Command Tutorials! These tutorials should help you to map specific Flux commands
 with your use case, and then see detailed usage. 
 
- - ``flux proxy`` (:ref:`ssh-across-clusters`): "Send commands to a flux instance across clusters using ssh"
  - ``flux-mini-submit`` (:ref:`flux-mini-submit`): "Submit a job in a Flux instance"
+ - ``flux proxy`` (:ref:`flux-proxy-command`): "flux proxy basics"
```
vsoch, line 10: Can we here give the reader a better sense of why they would want to use flux proxy, and how it differs from the "ssh across clusters" tutorial?

tutorials/commands/flux-proxy-command.rst
```diff
@@ -0,0 +1,194 @@
+.. _flux-proxy-command:
+
+=================
+Flux Proxy Basics
+=================
+
+With Flux, it is very common to create a Flux :ref:`subinstances<subinstance>`.  However, it can be confusing how to interact with those subinstances.
```
vsoch, line 7: So this is saying when I launch a flux job that controls some set of subinstances, I would use flux proxy to communicate with them? Or could I have flux controlling multiple clusters, and then flux proxy would give me access to a cluster? What is a subinstance (or what can it be) in concrete terms? Why would I want to communicate with it? E.g., what we might have here is "you might want to do this if X or Y or Z" is what you are doing.

tutorials/commands/flux-proxy-command.rst
```diff
@@ -0,0 +1,194 @@
...
+
+-----------------------
+Starting a Subinstance
+-----------------------
+
+First, let's create a subinstance to work with.  The most common way is through ``flux mini batch``.  We'll launch a subinstance and run a few jobs under that instance, we then call ``flux job status`` to wait on those jobs to finish.
+
+.. code-block:: sh
+
+   #!/bin/sh
```
vsoch, line 19: Wait where does flux mini batch come in? This looks like it's running 4x flux mini submit?

tutorials/commands/flux-proxy-command.rst
```diff
@@ -0,0 +1,194 @@
...
+
+   jobid1=`flux mini submit -n1 sleep inf`
+   jobid2=`flux mini submit -n1 sleep inf`
+   jobid3=`flux mini submit -n1 sleep inf`
+   jobid4=`flux mini submit -n1 sleep inf`
+   flux job status ${jobid1} ${jobid2} ${jobid3} ${jobid4}
+
+.. code-block:: console
+
+   > flux mini batch -n4 ./subinstance-jobs.sh
```
vsoch, line 30: I totally don't understand what is going on here, lol (I've never used batch before).

vsoch: OK going to stop here and submit this early feedback because I can't follow along (sorry, not familiar with batch!)

chu11: lemme try to add a bit more details then.  one issue is we are writing these tutorials out of order.  I jumped on `flux proxy` mostly b/c you already wrote the ssh-across-clusters one. Really `flux mini batch` probably should have come after `flux mini submit`.  which I will add to the todo list.

tutorials/commands/flux-proxy-command.rst
grondo: Maybe a better intro here would be to talk about traditional batch jobs running a batch script, and how in Flux those "batch jobs" are actually running a full instance of Flux, and the `flux proxy` command can be used to execute processes, including an interactive shell, as if they were connected directly to the instance of Flux running your batch workload.

Also, `flux proxy` is not restricted to running against subinstances. You can also use it to connect to a system instance running on another cluster, a test instance running locally, or as you show below, a Flux instance running under a Slurm, LSF, or other resource manager job.

tutorials/commands/flux-proxy-command.rst
grondo: To answer your question @v, `flux mini batch` is the `flux mini` utility that submits a "batch job" which in Flux is a job that runs an instance of Flux on the assigned resources with a batch script as its "initial program"

tutorials/commands/flux-proxy-command.rst
grondo: shouldn't we encourage the use of `--cc` in this case? Possibly using waitable jobs and  `flux job wait` for more "standard" practice.

tutorials/commands/flux-proxy-command.rst
```diff
@@ -0,0 +1,194 @@
...
+         fh2fcK1 achu     sleep       R      1      1   1.502m opal186
+         fa6txwZ achu     sleep       R      1      1   1.506m opal186
+
+But how could we submit a job to that subinstance?
+	 
+-----------------------
+Connect to the Instance
+-----------------------
+
+The easiest way to operate with a subinstance is to use ``flux-proxy``.  You can pass it the subinstance jobid and the command you want to run against that subinstance.  Lets submit a job via ``flux mini submit`` and list it via ``flux jobs``.
```
grondo, line 65: I'd remove subinstance here, might be less confusing to just say jobid.

tutorials/commands/flux-proxy-command.rst
```diff
@@ -0,0 +1,194 @@
...
+
+    > flux proxy f2NRGXaP flux jobs
+            JOBID USER     NAME       ST NTASKS NNODES     TIME INFO
+        f2sW9cPyZ achu     hostname    S      1      -        - 
+          fvhpz5V achu     sleep       R      1      1   4.143m opal186
+          foyvExo achu     sleep       R      1      1   4.148m opal186
+          fh2fcK1 achu     sleep       R      1      1   4.152m opal186
+          fa6txwZ achu     sleep       R      1      1   4.157m opal186
+
+Notice that we're running ``flux jobs`` against the subinstance itself.  We don't have to specify the ``--recursive`` option to list the jobids anymore.
```
grondo, line 80: More interesting might be to submit more work to the instance (this is where `flux job wait --all` would come in handy)

tutorials/commands/flux-proxy-command.rst
```diff
@@ -0,0 +1,194 @@
...
+
+You can also ssh to the node if you know the subinstance's Flux :ref:`URI<URI>`.  This may be useful if you want to proxy to the subinstance from outside of the cluster at a later time (see :ref:`SSH Across Clusters`<ssh-across-clusters>` for more information).
```
grondo, line 82: I think you mean here "you can also specify a _native_ uri" instead of a target jobid?

tutorials/commands/flux-proxy-command.rst
```diff
@@ -0,0 +1,194 @@
...
+    > flux proxy ssh://catalyst160/var/tmp/achu/flux-IVK7W4/local-0 flux jobs
+            JOBID USER     NAME       ST NTASKS NNODES     TIME INFO
+        f2sW9cPyZ achu     hostname    S      1      -        - 
+          fvhpz5V achu     sleep       R      1      1   6.371m opal186
+          foyvExo achu     sleep       R      1      1   6.376m opal186
+          fh2fcK1 achu     sleep       R      1      1    6.38m opal186
+          fa6txwZ achu     sleep       R      1      1   6.385m opal186
+
+
+If you do not specify a command to run, you will be dropped into a shell that will forward all commands to the subinstance.
```
grondo, line 98:
```suggestion
If you do not specify a command to run, you will be dropped into a shell that will forward all messages to the subinstance.
```
?

grondo (commented):
Also added some quick comments, though I didn't get all the way through.

tutorials/commands/flux-proxy-command.rst
chu11: for some of these small examples I figure its easier to do this than have to explain `--cc` (leaving `--cc` to the #195).  

Do we have a preference on `flux job wait` vs `flux job status` for something this small?

tutorials/commands/flux-proxy-command.rst
vsoch: okay - so you give it a script with other flux submit commands? As opposed to (what I've seen for other managers) a text file with some list of single commands (without the job manager directive) to run?

vsoch: How does that (conceptually) compare to a job array?

tutorials/commands/flux-proxy-command.rst
chu11: > Maybe a better intro here would be to talk about traditional batch jobs running a batch script, and how in Flux those "batch jobs" are actually running a full instance of Flux, and the flux proxy command can be used to execute processes, including an interactive shell, as if they were connected directly to the instance of Flux running your batch workload.

The way I imagined it was that we'd eventually have this discussed more in a `flux mini batch` tutorial.  Although I can probably go into it a little more than I have so the reader isn't quite as clueless.

tutorials/commands/flux-proxy-command.rst
grondo: Good points. Given that I'm fine with current example, was more of a question anyway.

tutorials/commands/flux-proxy-command.rst
chu11: > Wait where does flux mini batch come in? This looks like it's running 4x flux mini submit?

I listed the script and the run the `flux mini batch` right below, I guess I should put a sentence in between, perhaps its not obvious when skimming.

chu11:
Re-pushed, re-working the flow of the tutorial in a different way.  Instead of starting with `flux mini batch` I start with `flux mini alloc` and try to add a little more description about subinstances along the way.  I could certainly add a bit more, but I think its best to save that for an eventual `flux mini alloc` / `flux mini batch` tutorial that we can eventually point to.

chu11:
re-pushed with some minor tweaks given discussion from other PRs.  If there is still a little confusion of what is `flux mini batch` doing and why would I want to make a subinstance, I think that there will be future tutorials to deal with that.

tutorials/commands/index.rst
```diff
@@ -7,6 +7,7 @@ Welcome to the Command Tutorials! These tutorials should help you to map specifi
 with your use case, and then see detailed usage. 
 
  - ``flux mini submit/flux mini run`` (:ref:`flux-mini-submit`): "Submit a job in a Flux instance"
+ - ``flux proxy`` (:ref:`flux-proxy-command`): "Send commands to other Flux instances"
```
vsoch, line 10: Yeah I guess this is good for now - some of these are definitely more advanced.

Issue #193, opened 2023-02-07
Command Tutorials to do


vsoch:
- [ ] `flux proxy`: "I want to connect to a flux instance across clusters with ssh #192 
- [ ]  `flux proxy`: "send commands to a flux instance you've started"
- [ ]  `flux start`: "I want to start my own flux allocation/instance to launch jobs"
- [ ] `flux mini submit/run`: "I want to run or submit a job to a flux instance"

Please add to this list as you see fit!

chu11:
side note on the submitting jobs faster one, as it crosses multiple commands and maybe is more about techniques, may cross the threshold into general "tutorial".

vsoch:
@chu11 I think that might fit well for more of an "advanced tutorial" - to submit jobs more efficiently I imagine the user is pretty good at Flux but wants to be able to optimize that. We have entire sections of the docs just for Jobs so maybe as another page there? https://flux-framework.readthedocs.io/en/latest/jobs/index.html

Any part of the site we can also talk about in terms of a total re-organization - I'm not wed to anything and wanting to make it all better!

grondo:
One thing I've had in the back of my mind is that `flux mini bulksubmit` is generally inspired by GNU parallel, which has some great tutorials online. So there is a set of users out that aren't HPC users, but want to run things in parallel. Maybe a specific tutorial that is geared toward that class of user would be a nice intro. It could borrow from GNU parallel tutorials, but add on the benefits of having a resident resource manager handling all those jobs instead of a single process:
 * oh, you want to quickly list which jobs failed? Here you go (`flux jobs -f failed`)
 * oh, you want to look at the output of a failed job again? `flux job attach`
 * oh, you want to watch output as the jobs execute? `flux mini bulksubmit --watch --progress`
 * oh, you want to cancel something that hasn't run yet? `flux jobs + flux job cancel` or `flux pkill`
 * oh, you forgot a set of things and want to run them too? (just submit more work!)
 * oh, your workload needs more resources? All these examples work whether your Flux instance is running locally on your laptop with `flux start -s 1` or on a cluster of 1000 nodes?
 * oh, you only have access to a cluster with a traditional resource manager? That's ok, you can start a large Flux instance as a job under many resource managers.

And so on...

vsoch:
_oh, you want to do the Flux thing? Have no fear - Fluxman is here!_

[...]

PR #202, opened 2023-02-13
tutorials: add flux mini run tutorial/example

chu11:
originally in #193, we intended to be in the `flux mini submit` tutorial but I guess we forgot about it.  I didn't think it was worthwhile to separate it out to a new tutorial b/c its so small and highly related to `flux mini submit`.  So I put it in there.

Lots of mini cleanups along the way too.

### DIFF
--- a/tutorials/commands/flux-mini-submit.rst
+++ b/tutorials/commands/flux-mini-submit.rst
@@ -30,7 +30,7 @@ task on just one node with 2 cores allocated for the task.
 There are many different options to customize your job submission. For further
 details, please see :core:man1:`flux-mini`.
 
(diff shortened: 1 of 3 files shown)


cmoussa1 (approved):
LGTM!

chu11:
@cmoussa1 thanks for the review.  I jut realized that some of the cleanups I did in #200 conflict with this PR.  Since #200 is still being worked on, I just went ahead and stole those cleanups and put them in here so they wouldn't conflict with newer PRs I'm going to start.  Could you take a quick skim again and make sure these new cleanups are ok too.

cmoussa1 (approved):
The new cleanups LGTM too. Thanks @chu11!

PR #205, opened 2023-02-15
[wip] add example for submit and wait

vsoch:
I started discussing this example with @grondo in slack - and we should discuss the use cases for this particular example to distinguish when the user should use the FluxExecutor as opposed to a handle + jobspec + flux.job.submit. A more advanced user is going to want to know why they should know about two ways to do the same thing, and if it's not the same thing, the pros and cons of the two approaches.

This is a WIP as it warrants further discussion!


vsoch:
Update: I think we should clean up the use of jobspec first (e.g., being able to submit with one) and then figure out the context here and finish these up. I am also re-designing the flux-cloud submit given the new Flux Operator Python SDK and trying to figure out if I can minimize the number of specs of things (e.g., use a jobspec or a piece of one over there).

PR #206, opened 2023-02-15
faqs: add entry on shell prompt

chu11:
This comes up once in awhile, and once in awhile I'm like "where's that prompt to cut and paste" .... so added a FAQ entry.


cmoussa1 (approved):
LGTM!

PR #209, opened 2023-02-16
examples: remove unrendering code

### LINKED ISSUE #207: python api example something not rendering properly
Just noticed this when skimming through a build and then noticed it is on the read the docs site.

image

I assume that last chunk should be rendered into something?

chu11:
Problem: Some code at the bottom of the job submit example is not rendering properly.

The code snippet is unnecessary, so remove it.

Fixes #207


chu11:
@vsoch thanks

PR #210, opened 2023-02-17
tutorials: Add a flux job cancel tutorial

chu11:


tutorials/commands/flux-job-cancel.rst
```diff
@@ -0,0 +1,114 @@
+.. _flux-job-cancel:
+.. _flux-job-cancelall:
```
vsoch, line 2: Just a high level note - Dan and I were trying to remember how to cancel the other day and the intuitive thing (that follows other tools) would be `flux job cancel --all`. Maybe we could deprecate `cancelall` or (so the intuitive one is available) just provide both?

tutorials/commands/flux-job-cancel.rst
```diff
@@ -0,0 +1,114 @@
...
+----------------------
+
+The basic way to cancel a job is through ``flux job cancel``.  All you have to do is specify
+the jobid on the command line.  Here is a simple example after submitting a job.
+
+.. code-block:: console
+
+    $ flux mini submit sleep 100
+    ƒh35Dh5qRyq
+
```
vsoch, line 22: Maybe between here we could add like:

```markdown
If you've job submit the job, the id will be readily available. If you need to see a list of your current jobs, you can use `flux jobs`.
```

tutorials/commands/flux-job-cancel.rst
```diff
@@ -0,0 +1,114 @@
...
+a little bit, we see that the job is now canceled in ``flux jobs`` (state is ``CA``).
+
+------------------------
+Cancelling a lot of jobs
+------------------------
+
+The special ``flux job cancelall`` command allows you to cancel many jobs.  Several options allow you to
+target a specific set of jobs.
+
+To start off, lets create 100 jobs that will sleep infinitely.  We will use the special ``--cc`` option
```
vsoch, line 49: oh no, sleeping beauty jobs!

vsoch: What does `--cc` mean?

tutorials/commands/flux-job-cancel.rst
```diff
@@ -0,0 +1,114 @@
...
+        ƒjTWS5m1 achu     sleep       R      1      1   8.860s corona212
+        ƒjTUx6Um achu     sleep       R      1      1   8.870s corona212
+        ƒjTUx6Uk achu     sleep       R      1      1   8.870s corona212
+        ƒjTUx6Uj achu     sleep       R      1      1   8.870s corona212
+        ƒjTUx6Ui achu     sleep       R      1      1   8.871s corona212
+	<snip - there are many jobs running>
+
+As you can see, we have a lot of jobs waiting to run (state ``S``) and a lot of running jobs (state ``R``).
+
+Lets first ``flux job cancell`` without any options.
```
vsoch, line 74:
```suggestion
Lets first ``flux job cancel`` without any options.
```

vsoch: Or this should be "cancelall" ?

tutorials/commands/flux-job-cancel.rst
```diff
@@ -0,0 +1,114 @@
...
+
+.. code-block:: console
+
+    $ flux job cancelall
+    flux-job: Command matched 100 jobs (-f to confirm)
```
vsoch, line 79: I would expect this to be `flux job cancel --all --force` for some future update - I could guess that easily!

tutorials/commands/flux-job-cancel.rst
```diff
@@ -0,0 +1,114 @@
...
+    $ flux job cancelall -f
+    flux-job: Canceled 100 jobs (0 errors)
+
+    $ flux jobs
+               JOBID USER     NAME       ST NTASKS NNODES     TIME INFO
+
+As you can see all the jobs are now canceled after passing the ``-f`` option to ``flux job cancelall``.
+
+``flux job cancellall`` has several options to filter the jobs to cancel.  Perhaps the most commonly used
+option is ``-S``.  The ``-S`` option specifies the state of a job to cancel.  The most
```
vsoch, line 95: Does this have an equivalent long form? It might be easier to read if it's like `--state` and then the user would remember that (and be able to guess the short form) vs. learning to use `-S` but not knowing what it means.

tutorials/commands/flux-job-cancel.rst
chu11: I didn't happen to write `cancelall`, although I suspect it was because `flux job cancel` takes jobids as input while `flux job cancelall` does not.  So separating them out made more sense than having to deal with mixing up the logic of both together.  

But it's a fair point, perhaps bring up the `--all` option for a discussion in `flux-core`.

tutorials/commands/flux-job-cancel.rst
chu11: i assume "carbon copy", which we can mention (and hopefully its obvious what it does given what its called)

tutorials/commands/flux-job-cancel.rst
grondo: I think those commands were modeled after Unix/Linux `kill` vs `killall`. 

Considering we're still building interfaces, this command could be considered a "plumbing" command in the future, and so I wonder if a whole tutorial on `flux job cancel` is necessary. 
Feels like a "canceling jobs" howto might be better?

tutorials/commands/flux-job-cancel.rst
```diff
@@ -0,0 +1,119 @@
...
+
+Note that in this particular example we happened to know the jobid of our job.  If you do
+not know the the jobid of your job, you can always use ``flux jobs`` to see a list of all
+your currently active jobs.
+
+------------------------
+Cancelling a lot of jobs
+------------------------
+
+The special ``flux job cancelall`` command allows you to cancel many jobs.  Several options allow you to
```
grondo, line 50: This might be misleading since `flux job cancel` can also cancel many jobs.

chu11: oh good point, multiple jobids vs filtering mechanism.

tutorials/commands/flux-job-cancel.rst
```diff
@@ -0,0 +1,119 @@
...
+    $ flux job cancelall --states=pending
+    flux-job: Command matched 52 jobs (-f to confirm)
+
+    $ flux job cancelall --states=running
+    flux-job: Command matched 48 jobs (-f to confirm)
+
+As you can see ``flux job cancelall --states=pending`` would target the 52 pending jobs for cancellation and
+``flux job cancelall --states=running`` would target the current 48 running jobs for cancellation.
+
+And that's it! If you have any questions, please
```
grondo, line 118: Since cancelall is mentioned, might as well mention `flux pkill`?

chu11: i thought about that, but `pkill` is really about signaling.  So its a bit different?

tutorials/commands/flux-job-cancel.rst
vsoch: I think for the time being, in that `flux job cancel` is a command (and one we badly needed yesterday!) it should be included here and we can move it / change our minds about it when the CLI is refactored.

With respect to that, I have some pretty strong opinions about cli design so I hope you invite me to the conversation, of course bringing a silver spike and garlic in case I get out of hand :vampire:

tutorials/commands/flux-job-cancel.rst
vsoch: I didn't know Flux resorted to violence. :scream:

tutorials/commands/flux-job-cancel.rst
grondo: We also just merged `flux job last` which might be even easier than `flux jobs` in this case

chu11:
re-pushed with tweaks given the comments above, thanks!

tutorials/commands/flux-job-cancel.rst
chu11: In this particular case I was mostly trying to show that the job was running.  Since `flux mini submit` already output the jobid and then afterwards it showed the job was canceled.

tutorials/commands/flux-job-cancel.rst
```diff
@@ -0,0 +1,122 @@
...
+The ``flux job cancelall`` command allows you to cancel jobs without specifying jobids.
+By default it cancels all of your active jobs, but several options allow you to target a subset of the jobs.
+
+To start off, lets create 100 jobs that will sleep infinitely.  We will use the special ``--cc`` (carbon copy)
+option to ``flux mini submit`` that will submit 100 duplicate copies of the ``sleep`` job.
+
+.. code-block:: console
+
+    $ flux mini submit --cc=1-100 sleep inf
+    <snip - many job ids printed out>
```
grondo, line 62: Optional: could use --quiet to suppress jobids on standard output

tutorials/commands/flux-job-cancel.rst
grondo: No `pkill` cancels jobs

tutorials/commands/flux-job-cancel.rst
chu11: yeah, I didn't use `--quiet` in the #195 fast job submission guide too.  I just didn't want to have to explain another option I used :P

maybe it's not a net win??

grondo: Yeah, I only mentioned it since you had to snip the output. Wondering how that will work if output is ever autogenerated. 

I didn't get a chance to review the other tutorial pr

tutorials/commands/flux-job-cancel.rst
chu11: well crud, i guess i was just thinking `pkill(1)` ... yeah, i guess it should go there too

chu11:
ok, re-pushed adding a `flux pkill` section.

chu11:
re-pushed adding a note given work on https://github.com/flux-framework/flux-core/pull/5055

PR #212, opened 2023-02-26
deprecate flux-mini

garlick:
Here's a batch of updates to change references to the flux-mini commands to their new names which will appear in flux-core v0.39.0.

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -67,7 +67,7 @@ F58 to another using the :core:man1:`flux-job` ``id`` subcommand, e.g.
 
 .. code-block:: sh
 
-   $ flux mini submit sleep 3600 | flux job id --to=words
+   $ flux submit sleep 3600 | flux job id --to=words
    airline-alibi-index--tuna-maximum-adam
    $ flux job cancel airline-alibi-index--tuna-maximum-adam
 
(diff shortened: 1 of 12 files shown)


grondo:
I think as part of the new process here you have to remake the examples output locally before pushing:

```
  [2023-02-26T19:13:27.010Z] Testing that you built the auto-examples, if this fails run flux start make html locally.
  
  [2023-02-26T19:13:27.042Z] 🐨️ Checking for differences between /tmp/auto_examples and ./auto_examples
  
  [2023-02-26T19:13:27.043Z] Traceback (most recent call last):
    File ".github/scripts/check_diff.py", line 77, in <module>
      main()
    File ".github/scripts/check_diff.py", line 69, in main
      sys.exit(
  NameError: name 'sys' is not defined
```

It also looks like there's a bug in the `check_diff.py` script, but I think what it is trying to say is you have to run `make html` under `flux start` and commit the results and try again.

grondo (approved):
Thanks for doing this! Besides the auto-examples generation, this LGTM

vsoch (approved):

vsoch:
> I think as part of the new process here you have to remake the examples output locally before pushing:

That's correct! If you want to add sys feel free to do that too.  You'll want to use `flux start make html` because the examples need flux. https://github.com/flux-framework/flux-docs#setup

garlick:
I need to get my devel environment sorted - meanwhile  I just dropped the examples update (it was only one line)

vsoch:
That seems good to me! Can you point me to the example that needs an update? I can do that.

As a side note - it was working on these tutorials that gave me pause in the first place and my inner dialogue was like "We need to fix the commands first before teaching them!!"

garlick:
Thanks!  This one:
```
 garlick@system76-pc:~/proj/flux-docs/examples$ git grep mini
example_job_submit_api.py:# flux mini submit --ntasks=4 --nodes=2 --cores-per-task=2 sleep 10
```

vsoch:
See #213

PR #213, opened 2023-02-26
examples: remove reference to flux mini

vsoch:
And re-render the examples with that update. This is done via

```
flux start make html
```
in the container. I'm thinking we want to adjust the user as we did for flux core too, can do that in another PR.


garlick (approved):
Thanks!

PR #214, opened 2023-02-26
docker dev environment fix permissions

vsoch:
Problem: permissions are written as root in the current devcontainer
Solution: run them as the vscode user with uid/gid 1000

This is the same fix applied to the flux-core container - I also tested it out (and doing things with sudo) and it seems good!


garlick (approved):
Thanks!

PR #216, opened 2023-03-01
flux-accounting guide: add setup instructions for new flux-accounting service

### LINKED ISSUE #215: flux-accounting: update set up and installation instructions
Now that the flux-accounting service can be controlled with systemd, the installation and set up instructions should be updated to include this.

cmoussa1:
#### Background

flux-accounting has a new service that accepts `flux account` commands that interact with the flux-accounting database and can be controlled with `systemd`, but there are no setup instructions on how to get it going in the flux-accounting guide on our docs site.

---

This PR adds a couple of small fixes along with a set of notes on installing and setting up the flux-accounting service in the flux-accounting guide. Since this new addition to documentation will be required _as a result of_ the new flux-accounting release coming up in a few days, I will leave this is as [WIP] until the release has been finalized. Feel free to take a look at my changes proposed here in the meantime, though, in case there is any feedback. 👍

Fixes #215

### DIFF
--- a/guides/accounting-guide.rst
+++ b/guides/accounting-guide.rst
@@ -11,7 +11,7 @@ Flux Accounting Guide
     documented in this guide may change with regularity.
 
     This document is in DRAFT form and currently applies to flux-accounting
-    version 0.18.1.
+    version 0.31.0.
 
 ********
 Overview
--- a/guides/accounting-guide.rst
+++ b/guides/accounting-guide.rst
@@ -61,7 +61,27 @@ Installing Software Packages
 ============================
 
 The ``flux-accounting`` package should be installed on the management node
-from your Linux distribution package manager.
+from your Linux distribution package manager. Once installed, the service
+that accepts ``flux account`` commands and interacts with the flux-accounting
+database can be started.
+
+You can enable the service with ``systemctl``; if not configured with a custom
+path, the flux-accounting systemd unit file will be installed to the same
+location as flux-core's systemd unit file:
+
+.. code-block:: console
+
+  $ sudo systemctl enable flux-accounting
+
+The service can then be controlled with ``systemd``. To utilize the service,
+the following prerequisites must be met:
+
+1. A flux-accounting database has been created with ``flux account create-db``.
+The service establishes a connection with the database in order to read from
+and write to it.
+
+2. An active Flux system instance is running. The flux-accounting service will
+only run after the system instance is started.
 
 Accounting Database Creation
 ============================


vsoch:
I added a few notes to the flux-operator multi-tenant tutorial I think were helpful for accounting, namely:

- creating the database `flux account create-db`
- adding root bank `flux account add-bank root 1`
- adding user bank `flux account add-bank --parent-bank=root user_bank 1`
- adding users `flux account add-user --username=peenut --bank=user_bank`
- also showing how to add the user (to the system, probably first) e.g., `adduser`
- showing how to get the status of a user with `view-user` flux account view-user greatuser`

And a note about database ownership / permissions if it's not there yet! The complete guide with the examples above starts here: https://flux-framework.org/flux-operator/getting_started/tutorials/multi-tenancy.html#view-logs maybe it would make sense to add a few of these points? At least I think I checked that guide when I was making this tutorial and I came to the slack to bother everyone still with questions :laughing:

cmoussa1:
@vsoch good suggestions! I think that would also be very helpful to add some of these pointers to the flux-accounting guide. I think adding these might be a little out of scope for this PR, but I'd be happy to open a separate issue and a near-future PR to add it. Does that sound okay?

With that, now that flux-accounting `v0.22.0` has landed, I'll take this out of [WIP] and mark it as ready for review whenever convenient.

vsoch:
Yep!

guides/accounting-guide.rst
```diff
@@ -61,7 +61,27 @@ Installing Software Packages
...
+that accepts ``flux account`` commands and interacts with the flux-accounting
+database can be started.
+
+You can enable the service with ``systemctl``; if not configured with a custom
+path, the flux-accounting systemd unit file will be installed to the same
+location as flux-core's systemd unit file:
+
+.. code-block:: console
+
+  $ sudo systemctl enable /path/to/flux-accounting.service
```
garlick, line 74: The path should not be necessary here if the unit file is installed in the right place, which will be the case when installing from packages.   `sudo systemctl enable flux-accounting` should be sufficient.

guides/accounting-guide.rst
```diff
@@ -61,7 +61,27 @@ Installing Software Packages
...
+
+The service can then be controlled with ``systemd``. The service makes the
+following assumptions:
+
+1. A flux-accounting database has been created with ``flux account create-db``.
+The service establishes a connection with the database in order to read from
+and write to it.
+
+2. An active Flux system instance is running. The flux-accounting service will
+only run after the system instance is started.
```
garlick, line 84: Rather than characterize these as assumptions, maybe "prerequisites"?

guides/accounting-guide.rst
```diff
@@ -11,7 +11,7 @@ Flux Accounting Guide
     documented in this guide may change with regularity.
 
     This document is in DRAFT form and currently applies to flux-accounting
-    version 0.18.1.
+    version 0.30.0.
```
garlick, line 14: Needs to be bumped again.

Suggestion for later: move this document to the flux accounting project like we did with the admin guide and flux core.   and stop mentioning the version in the guide, since it's then more obvious that the guide applies to the current version.

(Side benefit: if you set up readthedocs infrastructure in flux-accounting then man pages for the commands would be easy to add)

garlick (approved):
Oof, so sorry this sat around for so long!

guides/accounting-guide.rst
cmoussa1: This is a good suggestion, thanks! I'll open an issue to move this document to flux-accounting 👍

cmoussa1:
Thanks for the review @garlick! I just pushed up some fixes based on your suggestions and opened an issue in flux-accounting to go ahead and move this document in the near future. Since you approved already, I can set MWP here shortly :-)

PR #220, opened 2023-03-28
waiting-for-jobs: add new guide

chu11:
Add a new guide on how to wait for jobs to complete.


jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+
+-----------------
+The --wait option
+-----------------
+
+The most basic way to wait for a job to complete on a submitted job is the ``--wait`` option on ``flux job submit``.  Simply put, if the ``--wait`` option is passed to ``flux job submit``, the command will not return until the job has completed.
+
+.. code-block:: console
+
+    $ flux submit --wait -n1 bash -c "sleep 30; /bin/false"
```
vsoch, line 22: I did not know this! TIL

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+    ƒEMds3VemM
+    <wait for command to finish>
+    $ echo $?
+    1
+
+The above command submits a job that simply sleeps for 30 seconds on one processor (``-n1``) and then runs ``/bin/false``.  The :ref:`jobid <fluid>` is immediately output, but the command won't return until the 30 second job has completed.
```
vsoch, line 28:
```suggestion
The above command submits a job that simply sleeps for 30 seconds on one processor (``-n1``) and then runs ``/bin/false``.  The :ref:`jobid <fluxid>` is immediately output, but the command won't return until the 30 second job has completed.
```

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+
+After the command has finished we print the exit code from ``flux submit``.  You'll notice the exit code is ``1``, which is the final exit code of the job, which in this case was ``1`` because we ran ``/bin/false``.
```
vsoch, line 30:
```suggestion
After the command has finished we print the exit code from ``flux submit``, which is  ``1``, because we ran ``/bin/false``.
```

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+
+---------------
+Flux Job Status
+---------------
+
+In most cases, you do not want to sit and wait for the current job submission to complete.  You would like to do other things, such as submit more jobs, and then wait for those specific jobs to complete.
```
vsoch, line 36: Indeed I don't! I have avocados to eat! Mountains to climb!

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+
+The ``flux job status`` command is the most basic way to wait for a specific job, based on jobid, to complete.  Pass it one or more jobids to wait on, and ``flux job status`` will return once all of the jobs have completed.  It will exit with largest exit code from any of the jobids specified.  If the job(s) have already completed, ``flux job status`` returns immediately.  It can be run as many times as the user would like against the same jobid.
```
vsoch, line 38: Is the context here that I've submit a bunch, and then (after that) I want to wait for a specific job?

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+
+You'll notice two things about this example.  First, the command returns immediately.  This is because the two jobs have already completed.  Second, the exit code is ``1``, which is the largest exit code of the two jobs passed to ``flux job status`` (one ran ``/bin/true`` and the other ran ``/bin/false``).
+
+-------------
+Flux Job Wait
+-------------
+
+``flux job wait`` behaves similarly to ``flux job status`` but there are some differences in using it, which come with pros and cons.
+
+The most notable difference is that in order to use ``flux job wait``, jobs must be passed the ``waitable`` flag.  Any job that is not passed the ``waitable`` flag will not work with ``flux job wait``.  In addition, the ``waitable`` flag can only be used in user Flux instances (i.e. non-system instances).  User Flux instances are usually started via ``flux alloc`` or ``flux batch``.
```
vsoch, line 80: How did this design come to be?

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+    $ flux job wait ƒ4btMovw
+    <we wait a little bit waiting for the job to finish>
+    $ echo $?
+    0
+
+Note that when submitting the job, we submitted it with the ``waitable`` flag via ``--flags waitable``.
+
+This doesn't really show us anything special, it seems to be the same as ``flux job status``.  Lets now look at the major advantages of ``flux job wait``.
+
+Perhaps the biggest advantage of ``flux job wait`` is that apriori knowledge of jobids is not necessary.  If ``flux job wait`` is specfied without any jobids, it will wait for the first job that completes amongst all of the jobs you have submitted via the ``waitable`` flag.
```
vsoch, line 97: So this is a flag that plops a job into a group you want to wait for? So it's appropriate for a scoped run of some kind?

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+    $ flux job wait
+    ƒ2Zjt9VSw
+    $ flux job wait
+    ƒ2XRcLY7Z
+    $ flux job wait
+    ƒ2WxyXSUF
+    $ flux job wait
+    flux-job: there are no more waitable jobs
+
+In this above example, we submit three jobs, sleeping for 60, 45, and 30 seconds respectively before running ``/bin/true``.  We then run ``flux job wait`` without any inputs.  You'll notice the jobids for the ``sleep 30`` job, then ``sleep 45`` job, then ``sleep 60`` job are returned in that order.  Finally, without any jobs left running with the ``waitable`` flag, ``flux job wait`` indicates there are no more waitable jobs.
```
vsoch, line 116: So it doesn't wait for all of them to complete (like the multiple one on the same line?) What is the use case for this if I have to run it a gazillion times?

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+
+.. code-block:: console
+
+    $ flux submit --flags waitable -n1 bash -c "sleep 60; /bin/true"
+    ƒ4YNPpFmAf
+    $ flux submit --flags waitable -n1 bash -c "sleep 45; /bin/true"
+    ƒ4YPufmCjq
+    $ flux submit --flags waitable -n1 bash -c "sleep 30; /bin/false"
+    ƒ4YSVQWfZq
+    $ flux job wait --all --verbose
```
vsoch, line 128: ohh this one makes sense! But what is the use case for without `--all`?

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+    ƒ4YSVQWfZq: task(s) exited with exit code 1
+    ƒ4YPufmCjq: job completed successfully
+    ƒ4YNPpFmAf: job completed successfully
+    $ echo $?
+    1
+
+This example is similar to the above, except one of the jobs runs ``/bin/false`` instead of ``/bin/true``.  When ``flux job wait --all`` is executed, you'll notice a message output indicating that one job has failed (the one that ran ``/bin/false``).  And similar to ``flux job status``, the exit code of ``1`` is returned due to the highest exit code of all the jobs.
+
+The biggest disadvantage of ``flux job wait`` compared to ``flux job status`` is that jobs can only waited on once.
```
vsoch, line 137:
```suggestion
The biggest disadvantage of ``flux job wait`` compared to ``flux job status`` is that jobs can only be waited on once.
```

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+
+    $ flux submit --flags waitable -n1 bash -c "sleep 30; /bin/true"
+    ƒBbk3qrdro
+    $ flux job wait ƒBbk3qrdro
```
vsoch, line 141: Why would I put the jobid at all? Wouldn't I just run `flux job wait` without any args like shown in the example above?

jobs/waiting-for-jobs.rst
```diff
@@ -0,0 +1,216 @@
...
+Here we've submitted yet another sleep job, and try to wait on the job twice with ``flux job wait``.  As you can see, an error is returned on the second attempt to wait on the job.
+
+You might be wondering, if you want to wait for a set of known jobids, is it better to use ``flux jobs status`` or ``flux job wait``?  Generally speaking, ``flux job wait`` is faster and more efficient than ``flux job status``.  It is especially more efficient with the ``--all`` option, instead of passing in a large list of jobids to ``flux job status``.
+
+As summary conclusion, here are a list of the pros and cons of using ``flux job status`` vs ``flux job wait``.
+
+Pros:
+
+- ``flux job wait`` more efficient when waiting for a set of jobs
+- Jobids do not need to be specified to ``flux job wait``
```
vsoch, line 154: So maybe just take that part of the tutorial out - don't show giving a job id to flux job wait if that shouldn't be learned.

jobs/waiting-for-jobs.rst
chu11: did you typo something here?  When I grep I don't see a reference to "fluxid".

vsoch: I assumed "fluid" should be fluxid, but if "fluid" is correct my mistake!

jobs/waiting-for-jobs.rst
chu11: Yes, think I should mention something to that affect?

vsoch: Yes exactly - you read between the lines.

jobs/waiting-for-jobs.rst
chu11: To be honest I'm not sure, since I didn't write it.  My suspicion is a user requested something and we did it? :P  Perhaps @grondo or @garlick know this specific history?

jobs/waiting-for-jobs.rst
chu11: I believe the typical use case is a user wants to know when a job has finished and can do some type of post-processing on its results while the other jobs keep on running.  They don't care which one finishes first/next, they just need to know that one has finished.

jobs/waiting-for-jobs.rst
chu11: ahh you're correct for this specific case, they wouldn't need to.  Would it be clearer to not put in the jobid in this case?

jobs/waiting-for-jobs.rst
chu11: I see your point.  I'll mention it, but definitely stress it less.

jobs/waiting-for-jobs.rst
chu11: probably good to stick one sentence in there to note this common use case.

chu11:
pushed a fixup, tweaking a few things, adding a sentence here and there given comments above

jobs/waiting-for-jobs.rst
vsoch: So you couldn't use `flux job status` for that?

chu11: `flux job status` requires you to input all of the jobids and doesn't exit until all of the jobs finish, thus more inconvenient.

grondo:
Perhaps something should be said in here to the effect of: `flux job wait` implements semantics similar to `wait(2)` and `waitpid(2)` system calls or the `wait` POSIX shell command. It is much more efficient than `flux job status`, but can only be called once per waitable job since the wait status is "reaped", and it requires instance owner privileges.

If you need to wait for thousands of jobs efficiently, or need to wait for single jobs as they complete, then `flux job wait` with waitable jobs is probably the best solution.

jobs/waiting-for-jobs.rst
grondo: what if there was a previously submitted waitable job that was not yet reaped? `flux job wait` doesn't necessarily only wait for the last submitted job...

jobs/waiting-for-jobs.rst
grondo: Only being able to wait on a job once is not necessarily a disadvantage, without it you would not be able to `flux job wait` in a loop (you'd just keep getting the same jobid continually). So there is a purpose here and each interface satisfies different se cases.  Instead of calling this a disadvantage, maybe the guide should discuss the use cases for which each interface is designed?

chu11:
re-pushed.  taking into account several of the comments above, re-worked the flow of the ``flux job wait`` section a bit.

chu11:
re-pushed, updating example script given completion of https://github.com/flux-framework/flux-core/issues/5033

chu11:
this should be updated to include `flux watch`

PR #225, opened 2023-03-30
WIP: add overview to understand subinstances

chu11:
This is re-work of my prior "flux mini batch/alloc" tutorial #208 where instead I call it a "understanding subinstances" overview.  It has some overview of my user instances/subinstances are useful and has some pictures (ehh not great ones, sorry, not great at it).  It then goes into illustration of subinstances using flux alloc/batch.  Those sections are largely the same from the prior PR #208.

One idea is that I scrap the "batch" part of this overview and have it "flow" into https://flux-framework.readthedocs.io/en/latest/jobs/batch.html.  But that section would have to be re-worked enhanced a bit first.


vsoch:
@chu11 in case they are helpful, here is the sequence of slides I'm using in my Kubecon talk to explain flux and instances / nesting. https://docs.google.com/presentation/d/1d05YmwTe0Rn1g8I5BLFYjYmR-FOZffQCjeLnOzyg93c/edit?usp=sharing

PR #226, opened 2023-03-30
fix bug with check_diff.py sys import missing

vsoch:
I think @chu11's simple PR failed because we are missing the sys import! Let's get this change in.

### DIFF
--- a/auto_examples/index.rst
+++ b/auto_examples/index.rst
@@ -30,7 +30,7 @@ of using the Flux Python API.
 .. only:: html
 
   .. image:: /auto_examples/images/thumb/sphx_glr_example_job_submit_api_thumb.png
-    :alt: Introductory example - Job Submit API
+    :alt:
 
   :ref:`sphx_glr_auto_examples_example_job_submit_api.py`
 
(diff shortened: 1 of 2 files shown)


auto_examples/index.rst
```diff
@@ -30,7 +30,7 @@ of using the Flux Python API.
 .. only:: html
 
   .. image:: /auto_examples/images/thumb/sphx_glr_example_job_submit_api_thumb.png
-    :alt: Introductory example - Job Submit API
+    :alt:
```
chu11, line 33: was this removal accidental?  seems out of place?

vsoch: Yeah I thought so too - let me poke around. What it probably means is that someone merged without updating the examples (and then missed this). It doesn't make sense it wouldn't be updated.

vsoch: It's just alt text for the image - I'm OK to remove it if we can't figure it out. It is a mystery though!

chu11: ok approved!

chu11 (approved):

auto_examples/index.rst
vsoch: Also - I should have stated this directly - the auto_examples directory is generated by the `flux start make html` as part of sphinx gallery, it's not generated by a human. The reason the other PRs are failing (aside from the sys import bug) is because it's finding this difference between what is rendered and what should be.

PR #228, opened 2023-04-26
expose flux comparison table

vsoch:
Problem: the comparison table is highly useful, but hidden under a section of the FAQ and hard to find.
Solution: add it as a navigation tab on the top (the rationale that some will come to the docs explicitly looking for it) along with as a link in the first line of the quickstart.

This has been a thought lingering on my mind for a few months, and after someone in the slack had the same trouble to validate my concern, I think we want to better expose our comparison table. For this PR, I've added it exposed in two places:

1. At the very top as its own navigation heading

image

We aren't best using these navigation headings (we only have two!) and I think a substantial number of users coming to the site are coming explicitly to answer the question "How does this Flux tool compare to..." so it should be at the top, front and center.

2. At the top of the quick start

For really the same use case, but if the person misses the top tab and clicks Contents -> quick start first.

image

That said - I'm not sure "Contents" is super meaningful - it's definitely correct but I wonder if there is a better term?

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -37,6 +37,13 @@ The framework consists of a suite of projects, tools, and libraries which may be
 
    projects
 
+.. toctree::
+   :maxdepth: 1
+   :caption: Comparison Table
+
+   tables/comparison-table
+
+
 Contributor Relevant RFCs
 -------------------------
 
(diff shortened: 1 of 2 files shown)


grondo (approved):
Seems good to me!

vsoch:
Thank you! And thank you for the review - I think these two PRs (that I pinged you on) are fairly low hanging fruit, and really important that we get out asap.

PR #230, opened 2023-04-26
improve content on front page

vsoch:
Problem: we got feedback that the frontpage of the docs is not super useful for providing information about flux, namely what makes it special.

Solution: add the flux "killer feature" and link to the comparison table

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -18,6 +18,8 @@ Welcome to Flux's documentation!
 
 Flux is a flexible framework for resource management, built for your site.
 The framework consists of a suite of projects, tools, and libraries which may be used to build site-custom resource managers for High Performance Computing centers.
+Unlike traditional resource managers, Flux can run as a parallel job under most launchers that support MPI, including under Flux itself. This not only makes batch scripts and workflows for Flux portable to other resource managers (just launch Flux as a job), but it also means that batch jobs have all the features of a full resource manager at their disposal, as if they have an entire cluster to themselves.
+If you are interested in a high level comparison of Flux to other resource managers, see :ref:`comparison-table`.
 
 .. toctree::
    :maxdepth: 2


index.rst
```diff
@@ -18,6 +18,11 @@ Welcome to Flux's documentation!
 
 Flux is a flexible framework for resource management, built for your site.
 The framework consists of a suite of projects, tools, and libraries which may be used to build site-custom resource managers for High Performance Computing centers.
+Flux distinguishes itself from other resource managers because it gives you a full resource management framework with your batch jobs, allowing 
+workflows and other scientific ensembles to schedule massive sets of jobs as if they have a cluster to themselves. In the same manner, you can create
+an entire cluster or "flux instance" that you control in a single container or in the cloud. Since Flux can also run as a job under other resource managers, 
+your workflow or batch job that uses Flux can run under Flux, Slurm, LSF, etc. If you are interested in a high level comparison of Flux to other
+resource managers, see :ref:`comparison-table`.
```
grondo, line 25: I wonder if we should describe more something like this (definitely rough drafting it all the way here)

Unlike traditional resource managers, Flux can run as a parallel job under most launchers that support MPI, including under Flux itself. This not only makes batch scripts and workflows for Flux portable to other resource managers (just launch Flux as a job), but it also means that batch jobs have all the features of a full resource manager at their disposal, as if they have an entire cluster to themselves.

etc. etc.

We could also mention that it is simple to test out flux locally via `flux start --test-size` which I think is another killer feature since new users or developers don't even have to have access to a cluster to explore and create with Flux.

Mayeb @garlick could also give some feedback here.

garlick: Well, flux's ability to run as a parallel job is definitely _my_ favorite killer feature :-)

My description would probably lean more towards the version proposed by @grondo,  Is it too much detail here though?  I think @vsoch has good intuition about these things so maybe she can give us her opinion?

vsoch: I like both of them - I can change the snippet to @grondo suggestion above.

vsoch:
ok updated!

garlick (approved):
This looks good to me!

vsoch:
@grondo how do I fix this? Looks like it fell off the merge train!

grondo:
@mergifyio refresh

grondo:
I don't know why mergify won't merge this one. Let's just merge it manually.

grondo:
Oh, I think it was because your branch was in this repo. We always work off forks in flux-framework.

vsoch:
Thank you! Sorry, neither of my approaches work here - I don't like adding forks to my personal account, so I use researchapps (or the org directly) and that preference is conflicting with mergebot here!

PR #231, opened 2023-04-27
Rabbit usage documentation

jameshcorbett:
Updates and simplifies the CORAL2 documentation to remove a lot of outdated configuration information. Also adds documentation about how to use rabbits.

The CORAL2 docs I don't expect anyone to look at really, except maybe I can point some admins to it occasionally. The rabbit page I expect to be used much more frequently.

### DIFF
--- a/tutorials/lab/coral2.rst
+++ b/tutorials/lab/coral2.rst
@@ -78,3 +78,24 @@ The lines should come before any call to load plugins.
 
 If Flux jobs that use Cray MPICH end up as a collection of singletons,
 that is usually a sign that Cray MPICH is trying to use libPALS.
+
+-----------------------------
+Configuring Flux with Rabbits
+-----------------------------
+
+In order for a Flux system instance to be able to allocate
+rabbit storage, the ``dws_jobtap.so`` plugin must be loaded.
+The plugin can be loaded in a  config file like so:
+
+.. code-block::
+
+    [job-manager]
+    plugins = [
+      { load = "dws-jobtap.so" }
+    ]
+
+Also, the ``flux-coral2-dws`` systemd service must be started
+on the same node as the rank 0 broker of the system instance
+(i.e. the management node). The ``flux`` user must have
+a kubeconfig file in its home directory granting it read
+and write access.
--- a/tutorials/lab/coral2.rst
+++ b/tutorials/lab/coral2.rst
@@ -1,13 +1,13 @@
 .. _coral2:
 
-==================================
-CORAL2: Native Flux on Cray Shasta
-==================================
+===========================
+CORAL2: Flux on Cray Shasta
+===========================
 
-The LLNL, LBNL, and ORNL next-generation systems like RZNevada, Perlmutter,
-El Capitan, and Frontier are in various stages of early access. They are
-similar in that they all use the HPE Cray Shasta platform, which requires
-a few additional components to integrate completely with Flux.
+The LLNL, LBNL, and ORNL systems like Tioga, Perlmutter,
+El Capitan, and Frontier are similar in that they all use the
+HPE Cray Shasta platform, which requires
+an additional component to integrate completely with Flux.
 
 --------------
 Things to Know
@@ -18,9 +18,7 @@ Things to Know
     Attempting to run further multi-node jobs will cause the excess jobs
     to fail. There is no limit on *submitted* multi-node jobs, and
     single-node jobs do not count towards the limit.
-#.  All nested Flux instances (e.g. instances created with ``flux batch``,
-    ``flux alloc``, or ``flux submit ... flux start``
-    should meet one of the following criteria:
+#.  All Flux instances should meet one of the following criteria:
 
     - Occupy a single node
     - Have exclusive access to the nodes they are running on (e.g. they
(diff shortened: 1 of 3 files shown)


tutorials/lab/coral2.rst
```diff
@@ -18,96 +18,82 @@ Things to Know
...
-    should meet one of the following criteria:
+#.  All Flux instances should meet one of the following criteria:
 
     - Occupy a single node
     - Have exclusive access to the nodes they are running on (e.g. they
       do not share their resources with sibling instances).
 
     Instances that do not meet one of the above criteria will not work properly.
 
+By default Flux reserves ports 11000-11999 for itself. At any given
```
vsoch, line 29: I didn't know this! What are the ports used for?

tutorials/lab/coral2.rst
```diff
@@ -18,96 +18,82 @@ Things to Know
...
-  tar -xzvf flux-sched-0.18.0.tar.gz && cd flux-sched-0.18.0
+Applications linked to Cray MPICH will work natively with Flux
+provided the Cray MPICH library uses the PMI2 protocol instead of
+the homespun Cray PMI and libPALS. For Flux to support libPALS,
+flux-coral2 must be built (see above) and Flux must be configured
+to offer libPALS support. This is done by setting the "pmi" shell
+option to include "cray-pals" on a per-job basis like so:
 
-  ./configure --prefix=$PREFIX && make -j && make install && cd ..
+``flux submit -n1 -opmi=cray-pals hostname``
```
vsoch, line 63: Should we put this in code block so it better stands out?
```suggestion
.. code-block:: console

    $ flux submit -n1 -opmi=cray-pals hostname
```

And a tiny detail I just noticed - I think a lot of the other guides put the `$` to indicate the console - we should decide either way (and then make sure they are all consistent!)

tutorials/lab/coral2.rst
```diff
@@ -18,96 +18,82 @@ Things to Know
...
-  in your environment and run with $PREFIX/bin/flux"
+    [job-manager]
+    plugins = [
+      { load = "dws-jobtap.so" }
+    ]
 
+Also, the ``flux-coral2-dws`` systemd service must be started
+on the same node as the rank 0 broker of the system instance
+(i.e. the management node). The ``flux`` user must have
+a kubeconfig file in its home directory granting it read
```
vsoch, line 98: kubeconfig like a kubernetes thing? Or something else?

tutorials/lab/rabbit.rst
```diff
@@ -0,0 +1,127 @@
...
+a string containing one or more DW directives, or to a list of
+singleton DW directives.
+
+DW directives are strings that start with ``#DW``. Directives
+that begin with ``#DW jobdw`` are for requesting storage that
+lasts the lifetime of the associated flux job. Directives that
+begin with ``#DW copy_in`` and ``#DW copy_out`` are for
+describing data movement to and from the rabbits, respectively.
+
+The usage is most easily understood by example.
```
vsoch, line 22: Shameless request - place a :rabbit: somewhere in here... :grin:

tutorials/lab/coral2.rst
```diff
@@ -18,96 +18,82 @@ Things to Know
...
+and write access.
```
vsoch, line 99: Should this end with something like "You can read more about Flux with rabbit storage here (link to the rabbit doc below?)

tutorials/lab/rabbit.rst
```diff
@@ -0,0 +1,127 @@
+.. _rabbit:
+
+=================
+Flux with Rabbits
+=================
+
+
```
vsoch, line 7: Maybe we could open with a sentence of two that says what a rabbit is, and example cases when I might want to use them (even a bullet list could work here!)

vsoch (commented):
These are great! I think updating the (likely now deprecated) docs for coral2 is badly needed. I added a few suggestions for both - a lot of them tiny nits and you can pick and choose what you think is addressing. I don't think I can be of much help reviewing the technical content (I've never used rabbits) but overall the description is very good.

tutorials/lab/coral2.rst
jameshcorbett: They're used by Cray PMI for communication, I believe.

tutorials/lab/coral2.rst
jameshcorbett: Yup! https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/

tutorials/lab/rabbit.rst
jameshcorbett: That's probably a good idea but I think I'm going to punt on it because (for now) all our users are going to be motivated to try things out and will probably know what they want to use the rabbits for. Besides I think all the examples I can cook up myself would get something wrong...

tutorials/lab/coral2.rst
jameshcorbett: I think it's OK without that transition. This document is for administrators, the other is for users.

tutorials/lab/coral2.rst
jameshcorbett: Good point, will fix.

bdevcich-hpe (approved):
`rabbit.rst` looks good to me

tutorials/lab/rabbit.rst
```diff
@@ -0,0 +1,128 @@
...
+	#!/bin/bash
+
+	#FLUX: -N 2
+	#FLUX: -q pdebug
+	#FLUX: --setattr=dw="""
+	#FLUX: #DW jobdw type=xfs capacity=1TiB name=my_xfs
+	#FLUX: #DW jobdw type=lustre capacity=10GiB name=my_lustre
+	#FLUX: """
+
+	echo "Hello World!" > $DW_JOB_my_lustre/world.txt
```
behlendorf, line 48: Shouldn't this line be dropped?

tutorials/lab/rabbit.rst
```diff
@@ -0,0 +1,128 @@
...
+	to batch scripts.
+
+Requesting a 10 gigabyte XFS file system on the command line with data movement
+both to and from the rabbits:
+
+.. code-block:: console
+
+	$ flux alloc -N2 --setattr=dw="#DW jobdw type=xfs capacity=10GiB name=project1
+	#DW copy_in source=/p/lustre1/behlendo/zeros_in destination=\$DW_JOB_project1/
+	#DW copy_out source=\$DW_JOB_project1/ destination=/p/lustre1/behlendo/zeros_out/"
```
behlendorf, line 111: How about this which is a little more generic.

```
$ flux alloc -N2 --setattr=dw="#DW jobdw type=xfs capacity=10GiB name=project1
#DW copy_in source=/p/lustre1/$USER/file_in destination=\$DW_JOB_project1/
#DW copy_out source=\$DW_JOB_project1/ destination=/p/lustre1/$USER/dir_out/"
```

While testing this out I also noticed I get the following error if the `/p/lustre1/$USER/file_in ` doesn't exist.  If it does things work as expected.

```
24.260s: job.exception type=exception severity=0 DWS/Rabbit interactions failed: workflow in 'Error' state too long: DW Directive 1: Staging operation failed
```

tutorials/lab/rabbit.rst
jameshcorbett: How come? I thought it couldn't hurt to show some extremely basic usage.

tutorials/lab/rabbit.rst
```diff
@@ -0,0 +1,128 @@
...
+begin with ``#DW copy_in`` and ``#DW copy_out`` are for
+describing data movement to and from the rabbits, respectively.
+
+The usage is most easily understood by example.
+
+
+Examples of jobdw directives
+----------------------------
+
+Requesting a 10 gigabyte XFS file system on the
```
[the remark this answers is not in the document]
jameshcorbett: Good idea. I do say so elsewhere, but I'll add that here too.

tutorials/lab/rabbit.rst
jameshcorbett: Don't want everyone trying to copy your lustre data eh? I'll fix, and add a note that the source must exist.

tutorials/lab/rabbit.rst
behlendorf: I'm not very familiar with Flux job scripts, but won't this run on the node submitting the job?  In which case `$DW_JOB_my_lustre/` won't be defined nor will the filesystem be available.

tutorials/lab/rabbit.rst
behlendorf: > and add a note that the source must exist.

Let's also open an issue for this.  This should probably be reported as a fatal error, not as a "it took too long so I gave up" kind of error.  As a user it's going to be very unclear what the problem was given that error.

tutorials/lab/rabbit.rst
jameshcorbett: No it should work just like a Slurm job script, so it will run in the job allocation.

tutorials/lab/rabbit.rst
jameshcorbett: I'll open an issue. But the problem is more general: there's no way for Flux to know the difference between a transient and a fatal error. They're reported the same way. So no matter what the error is, Flux will report it as "it took too long so I gave up"

behlendorf: Right, this is something HPEs software would need to return as a fatal error to Flux.  @bdevcich-hpe any thoughts on this?

tutorials/lab/rabbit.rst
behlendorf: Ah okay, thanks for clearing that up for me.

behlendorf (approved):

tutorials/lab/rabbit.rst
jameshcorbett: https://github.com/NearNodeFlash/NearNodeFlash.github.io/issues/46 and https://github.com/NearNodeFlash/NearNodeFlash.github.io/issues/47

jameshcorbett:
@garlick when you get a chance to look it over, I think the technical content in `rabbit.rst` is correct since both Brian and Blake read it over. But if you could look at `coral2.rst` that would be helpful.

garlick (approved):
LGTM!  This doc seems like it could use some further development, but this is a positive step, so full speed ahead as far as I'm concerned.

jameshcorbett:
Thanks! Setting MWP.

PR #238, opened 2023-05-03
admin guide: provide helpful debugging info for PAM limits

garlick:
@grondo and I both ran into shells printing:
```
flux-shell[0]:  WARN: rlimit: nproc exceeds current max, raising value to hard limit
```
and it took a little sleuthing to figure out what was going on.  Add a small trail of bread crumbs to the admin guide for next time.


grondo (approved):
LGTM! Only suggestion would be to add the exact error from the flux-shell for potential matching of search terms. Not a requirement though!

garlick:
Added the actual message.  I had never tried to include a code snippet in one of those note/warning boxes before and wasn't sure it was possible!  It is, you just indent.  Setting MWP.

PR #240, opened 2023-05-17
rabbit: note allowed chars in name field

jameshcorbett:
Problem: the allowed characters in the 'name' field of DW strings is very limited. Some users
may attempt illegal names, and the error message
they would get is unhelpful.

Add a note about the allowed characters in the
name field.


vsoch (approved):
This LGTM - someone else from Flux that knows about rabbits can give the final stamp of approval.

jameshcorbett:
> someone else from Flux that knows about rabbits

There isn't really anyone, it's just me. The information is correct (verified with HPE and in practice) so I'm just going to set MWP.

vsoch:
🙌🐰Hail!!

PR #249, opened 2023-07-31
update adminarch image to use flux batch

vsoch:
Problem: the adminarch.png is showing flux mini batch, which is deprecated.
Solution: edit the image so it says flux batch.

I'm not sure what a .dia file is, but it didn't open with Gimp or Inkscape, and it was easy enough to copy the color and move the batch over a smidge to fix the graphic!


garlick:
That's a source file for the [dia](https://wiki.gnome.org/Apps/Dia) diagram editor.   I can fix.

PR #250, opened 2023-07-31
admin: update command in diagram

garlick:
Problem: the opening diagram in the admin guide references flux mini batch, now deprecated.

Update image to reference flux batch instead.


grondo (approved):
LGTM!

PR #253, opened 2023-09-25
jobs: cleanup jobs/batch.rst

grondo:
Someone complained that when they searched "start Flux interactively" they got a reference to starting Flux under Slurm `salloc` in a section about batch jobs.

Since this didn't make much sense, I wanted to remove this section. I didn't want to go into too much cleanup here, so I just went through the `jobs.batch.rst` docs and removed erroneous or outdated information, and corrected an example.


jameshcorbett (approved):
LGTM!

grondo:
Thanks @jameshcorbett!

PR #259, opened 2024-01-19
admin-guide: clarify queue property name requirements

grondo:
Problem: There is a statement in the admin guide that it is good practice to use property names that match the corresponding queue name, but no reason for this practice is given.

Explain that queue properties do not need to match the queue name, but note that it is good practice because it allows flux-resource list to suppress these duplicate names in is output.


garlick (approved):
Thanks for that!

grondo:
Thanks!

PR #261, opened 2024-01-31
glossary: add expanded definition for 'fair-share'

cmoussa1:
#### background

flux-accounting doesn't have a clear definition for fair-share and its place in the priority plugin in terms of calculating a job's priority in the Accounting Guide.

---

This PR adds 'fair-share' as a term in the Glossary section with an expanded definition.

It also makes the "Glossary" section a reference-able one in the flux-accounting guide so that when fair-share is mentioned in the **Multi-Factor Priority Plugin** section, the expanded definition can be easily linked to.

~~_leaving as [WIP] for now because I'm having trouble getting Sphinx installed on my recently updated Mac and can't build it locally 😔_~~

### DIFF
--- a/guides/accounting-guide.rst
+++ b/guides/accounting-guide.rst
@@ -289,7 +289,8 @@ importance in the overall priority calculation. The current factors present in
 the multi-factor priority plugin are:
 
 * **fair-share**: the ratio between the amount of resources allocated vs. resources
-  consumed.
+  consumed. See the :ref:`Glossary definition <glossary-section>` for a more
+  detailed explanation of how fair-share is utilized within flux-accounting.
 
 * **urgency**: a user-controlled factor to prioritize their own jobs.
 
@@ -311,6 +312,7 @@ Both "types" of jobs, *running* and *active*, are based on Flux's definitions
 of job states_. *Active* jobs can be in any state but INACTIVE. *Running* jobs
 are jobs in both RUN and CLEANUP states.
 
+.. _glossary-section:
 
 ********
 Glossary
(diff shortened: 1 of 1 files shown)


vsoch (commented):
Seems good to me! But you mentioned the context it's used in flux is different than fair share as a scheduling algorithm? How is that?

cmoussa1:
Fair-share is only one factor (of multiple) that influences a job's priority when it is submitted. Fair-share is mentioned as just one factor (of multiple) in the **Multi-Factor Priority Plugin** section, just above the **Glossary**, but perhaps I could do a better job making this more clear. What do you think?

vsoch:
> Fair-share is only one factor (of multiple) that influences a job's priority when it is submitted. Fair-share is mentioned as just one factor (of multiple) in the Multi-Factor Priority Plugin section, just above the Glossary, but perhaps I could do a better job making this more clear. What do you think?

Yeah - I'm not sure the best way to do this, but I think we should / might give the reader definitions and then (maybe a second paragraph) explain the context for flux. Just reading the definition without that people would guess we use a fair share scheduling algorithm.

cmoussa1:
Sorry, I am probably just misunderstanding you. Is the **Multi-Factor Priority Plugin** section not sufficient for what you are envisioning? My thinking there was that section outlines all of the factors currently enabled, which mentions fair-share as just one factor and a link to an expanded definition of what it means in the **Glossary**.

grondo (approved):
LGTM!

cmoussa1:
Thanks! Setting MWP

PR #264, opened 2024-03-01
relocate the admin guide to flux core

garlick:
This should go in after flux-framework/flux-core#5763.


grondo (approved):
LGTM! I wonder if there is a way to redirect from the old url to the new location just in case? Probably not a big deal if not, we probably know all the user's of the admin guide thus far.

garlick:
I'll go ahead and set MWP.  I'm not too concerned about redirects on this one, but we can always figure out how to do that later if needed

PR #266, opened 2024-03-02
debugging: add instructions for debugging with DDT

grondo:
Problem: There's a known workaround for debugging Flux jobs with DDT, but this is not documented.

Add a section on debugging with DDT to the job/debugging section of the Flux docs. Reword the introduction to note that both TotalView and DDT have been tested with Flux.


grondo:
Thanks!

PR #270, opened 2024-04-27
coral2: add instructions for building rabbit JGF

jameshcorbett:
Problem: there is no documentation for how to configure Fluxion for use with rabbits.

Add a couple of sentences of detail to the coral2 page.


jameshcorbett:
I never set MWP on this! Doing that now.

PR #272, opened 2024-08-07
tutorials: make CORAL example consistent

wihobbs:
Problem: the CORAL interactive tutorial uses the username "elvis" in all places but one.

Solution: make the document consistent by removing the last reference to "hobbs17" and replacing with "elvis".

PR #273, opened 2024-08-09
[WIP] accounting guide: insert a dummy change

cmoussa1:
#### Problem

We're trying to figure out why the auto examples generation is acting up with a simple name change in one of the `.rst` files.

---

This PR makes a dummy change to one of the guides to see if the auto examples generation still fails. This PR should not be merged, but just to see if the auto examples generation in GitHub actions still fails.

### DIFF
--- a/guides/accounting-guide.rst
+++ b/guides/accounting-guide.rst
@@ -43,6 +43,8 @@ cluster only.  Sites would typically populate the database and keep it up to
 date automatically using information regularly pulled or pushed from an
 external source like an identity management system.
 
+inserting a dummy change here
+
 ******************************
 Installation and Configuration
 ******************************


cmoussa1:
@wihobbs looks like this PR failed too with the same error message:

```
😢️ Auto examples were not updated! Different size of index.rst:
  1497 vs. 1557
```

Perhaps we need to open an issue on this? I wonder if something is acting up with GitHub actions? 🤷

grondo:
@vsoch, any ideas?

wihobbs:
I think this would be fixed with a rebase since I pushed changes to address it in #272. In any case, @cmoussa1 want to close?

cmoussa1:
OK, I'll close here. Thanks

PR #274, opened 2024-08-15
faq: update openmpi faq to reference pmix

wihobbs:
I put and/or because it sounds like you need one on Azure (or possibly both) and the other on LC resources. Could differ at other centers too depending on if pmix is the default.

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -604,9 +604,11 @@ caused segfaults of MPI in ``MPI_Finalize`` when UCX PML was used.
 4.0.6+ or 4.1.1+.
 
 A special `job shell plugin <https://github.com/flux-framework/flux-pmix>`_,
-offered as a separate package, is required to bootstrap the upcoming openmpi
+offered as a separate package, is required to bootstrap the openmpi
 5.0.x releases.  Once installed, the plugin is activated by submitting a job
-with the ``-ompi=openmpi@5`` option.
+with the ``-opmi=pmix`` option. In fact, ``-o pmi=pmix`` also works for earlier 
+versions of OpenMPI, and appears to avoid a UCX related deadlock in OpenMPI 
+4.1.2 (see `flux-framework/flux-core#5460. <https://github.com/flux-framework/flux-core/issues/5460>`_)
 
 .. _openmpi_config:
 


garlick:
Oh good catch.  In fact, the `openmpi@5`option is no longer supported (hasn't been since 2021!)

You might want to tack on:

> In fact `-o pmi=pmix` also works for earlier versions of openmpi too, and appears to avoid a UCX related deadlock in openmpi 4.1.2 (see https://github.com/flux-framework/flux-core/issues/5460)

wihobbs:
Thanks @garlick! Added your changes verbatim, dropped `-ompi=openmpi@5`, rebased and pushed.

garlick:
Check the CI preview - looks like the issue link isn't rendering properly.  Also, maybe it would read better if the sentence I suggested is in the same paragraph as the other one?

faqs.rst
```diff
@@ -606,7 +606,11 @@ caused segfaults of MPI in ``MPI_Finalize`` when UCX PML was used.
 A special `job shell plugin <https://github.com/flux-framework/flux-pmix>`_,
 offered as a separate package, is required to bootstrap the upcoming openmpi
 5.0.x releases.  Once installed, the plugin is activated by submitting a job
-with the ``-ompi=openmpi@5`` option.
+with the ``-opmi=pmix`` option.
+
+In fact, ``-o pmi=pmix`` also works for earlier versions of OpenMPI, and
+appears to avoid a UCX related deadlock in OpenMPI 4.1.2 (see 
+`flux-framework/flux-core#5460 <https://github.com/flux-framework/flux-core/issues/5460>`)
```
vsoch, line 613: I think the link needs to end with an underscore:

```suggestion
`flux-framework/flux-core#5460 <https://github.com/flux-framework/flux-core/issues/5460>`_)
```

wihobbs: Yep, that was it. Forgot to push 🤦

wihobbs:
I can't get this to build locally because of (I think) lab network weirdness, so waiting for the RTD to render so I can check that it all looks right.

wihobbs:
@garlick I also dropped "upcoming" when referencing OpenMPI 5, since it was released July 22, 2024.

garlick (approved):
Thanks!

PR #275, opened 2024-08-21
coral2: link to HPE rabbit docs

jameshcorbett:
Problem: HPE has documentation about the usage of #DW directives that is much more detailed and formal than the documentation given in this repo.

Remove the duplicated docs and link to the HPE documentation.

@bdevcich @ajfloeder want to make sure this is OK with you.

PR #276, opened 2024-08-22
coral2: update HPE rabbit docs link

jameshcorbett:
Dumb change but pointing to 'latest' wasn't an option when #275 was merged, it only became an option when HPE published a new release two hours ago.


cmoussa1 (approved):
👍

PR #277, opened 2024-09-03
coral2: change rabbit docs to use -S flag

jameshcorbett:
Problem: flux-core PR github.com/flux-framework/flux-core/pull/6238 introduced the `-S` option as shorthand for `--setattr`, and users will most likely prefer using `-S`, but only `--setattr` is described in the documentation.

Change the documentation to list both options and make the examples use `-S`.

### DIFF
--- a/tutorials/lab/rabbit.rst
+++ b/tutorials/lab/rabbit.rst
@@ -11,8 +11,12 @@ How to Allocate Rabbit Storage
 Request rabbit storage allocations for a job
 by setting the ``.attributes.system.dw`` field in a jobspec to
 a string containing one or more DW directives. A JSON list of DW directives
-is also accepted, but cannot be provided on the command line (it is more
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

It is a style choice, and I'm sure you're already aware, but since `-S` requires an argument, no space is necessary between `-S` and `dw`, so
```
$ flux alloc -N2 -Sdw="#DW jobdw type=xfs capacity=10GiB name=project1
```
will work too.

Just pointing that out in case it wasn't obvious :-)

jameshcorbett:
Yeah I was thinking users might find it easier to read maybe?

grondo:
Yes, good point. I wasn't suggesting any changes, just making sure we didn't think a space between `-S` and its argument is necessary.

jameshcorbett:
Makes sense. Thanks for the review! Setting MWP.

PR #285, opened 2024-10-23
Fix broken links and add linkcheck back to CI

wihobbs:
Problem: there are some broken links in our docs.

Fix them.

### DIFF
--- a/jobs/batch.rst
+++ b/jobs/batch.rst
@@ -7,8 +7,11 @@ Batch Jobs
 In Flux, a batch job is a script submitted along
 with a request for resources. When resources become
 available, the script is run as the
-`initial program <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_8.html#initial-program-program-1>`_ of a new single-user instance of Flux,
-with the batch user as `instance owner <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_8.html#terminology>`_.
+`initial program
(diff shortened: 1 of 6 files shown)


jobs/batch.rst
```diff
@@ -6,9 +6,9 @@ Batch Jobs
 
 In Flux, a batch job is a script submitted along
 with a request for resources. When resources become
-available, the script is run as the
-`initial program <https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_8.html#initial-program-program-1>`_ of a new single-user instance of Flux,
```
jameshcorbett, line 10: https://flux-framework.readthedocs.io/projects/flux-core/en/latest/guide/glossary.html#term-initial-program

https://flux-framework.readthedocs.io/projects/flux-core/en/latest/guide/glossary.html#term-instance-owner

jameshcorbett (approved):
Looks reasonable, there are two links that just need to be changed I think rather than removed.

jameshcorbett:
Oh sorry didn't see this was a draft :|

wihobbs:
eh maybe it doesn't have to be. I'd like to get back to automating linkcheck with CI but maybe we can accomplish that in a separate pr? Thoughts, @jameshcorbett?

jameshcorbett:
It depends on how much work there is to be done. If you think you could do it without a ton more effort that would be nice, because the proper way to review this PR as it is now would be to manually verify that all the links work which is tedious

jameshcorbett (approved):
💯

wihobbs:
Thanks @jameshcorbett! I'll set MWP.

PR #291, opened 2024-12-05
rabbit: add docs on preset #DW strings

jameshcorbett:
Problem: flux-coral2 supports preset DW strings, but they aren't described in the rabbit docs.

Add documentation.

### DIFF
--- a/tutorials/lab/rabbit.rst
+++ b/tutorials/lab/rabbit.rst
@@ -18,6 +18,53 @@ creating jobspecs directly.)
 On the command line, set the ``.attributes.system.dw`` field by passing flags
 like ``-S dw="my_string"`` or ``--setattr=dw="my_string"``.
 
+The simplest way to use rabbits is to use preset DW strings that
+administrators may have configured on a system. For more flexibility,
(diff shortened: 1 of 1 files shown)


tutorials/lab/rabbit.rst
```diff
@@ -18,6 +18,53 @@ creating jobspecs directly.)
...
+Running
+
+.. code-block:: bash
+
+	flux config get rabbit.presets | jq .
+
+will display a mapping of the presets that have been configured on the
+current system. The keys of the mapping are the strings that you can
+pass to ``-S dw=...``, the values of the mapping are the actual DW directives
+they presets correspond to.
```
cmoussa1, line 38: Sorry if I am misunderstanding; `their` instead of `they` here?

cmoussa1 (approved):
LGTM! Nice add :-)

tutorials/lab/rabbit.rst
jameshcorbett: Good catch thanks! I think I wanted `the` actually (although `their` would also work). Fixed and force-pushed. Setting MWP

PR #297, opened 2025-04-08
tutorials: deprecate coral2 docs in favor of `flux-coral2`

jameshcorbett:


jameshcorbett:
I would remove them but I think there are external links to them and I don't want those to get 404s, at least not for a few months.

garlick:
There might be something to play with here:

https://readthedocs-theme.readthedocs.io/en/latest/user-defined-redirects.html

Not sure if that would work (and not sure why I didn't find it when I moved the admin guide!)

jameshcorbett:
Thanks for the pointer @garlick ! I set up some redirects on the readthedocs page and they seem to work so I removed the docs instead of adding a `warning` banner. Setting MWP.

PR #298, opened 2025-04-16
project: flux-coral2, make search-able

vsoch:
The headers do not seem to be searchable, so for a user to find the new link to flux-coral2 we need to add a little bit of description that includes the term. Since a user might also be looking for a system that is relevant, I added the primary and testing systems too. Here is what the search result looks like:

image

And the page:

image

I like this design better (with description) because many of the project names don't speak for themselves - we might consider adding brief descriptions to the others as well.


jameshcorbett (approved):
Looks great but some files were included accidentally it looks like?

vsoch:
I think we have those already:

https://github.com/flux-framework/flux-docs/tree/master/auto_examples

And when I built, the zip files were re-generated (with different timestamps?) I didn't do any git add or modifications to any other files aside from the projects one, and the clone was fresh.

jameshcorbett:
Yeah, those files are annoying. I don't really know what they are or how they're generated but I think if they're changed it should at least be a separate commit and give a reason why. I'd just drop them from the commit if I were you and ignore the issue for now. Otherwise worst case we might be changing them around all the time, in every PR.

vsoch:
It would be hard to decouple from an update because they are generated with make html. I think redoing the PR to get around a change here for a zip file date is excessive for a docs site.

vsoch:
When make html is run, it downloads new Python assets, so I suspect something changed in a dependency that updated those files. Notice that other auto generated files were not updated. We have to update the examples incrementally as they go into a static deployment, so those changes (in one changeset) would show up eventually. 🤷‍♀️

jameshcorbett:
Makes sense I'd say go ahead and set MWP!

vsoch:
Thanks @jameshcorbett - will do.

PR #301, opened 2025-05-12
contributing: update testing section tips

sam-maloney:
Problem: flux-framework/flux-core#6550 removed the `make check-prep` target, but it is still referenced in the testing tips.

Update the documention with the preferred `make check TESTS=` idiom. Also adds a another tip on running individual tests with `make`, and documents the flags specified in the existing tip.


grondo (approved):
Great contribution, thank you!

PR #306, opened 2025-08-26
improve FAQ entry about filling up `/tmp`

grondo:
@ryanday36 noticed a typo in the FAQ entry on using `statedir`. Additionally, it was suggested that adding some example errors would be helpful here.

### DIFF
--- a/faqs.rst
+++ b/faqs.rst
@@ -67,9 +67,9 @@ F58 to another using the :core:man1:`flux-job` ``id`` subcommand, e.g.
 
 .. code-block:: sh
 
-   $ flux submit sleep 3600 | flux job id --to=words
-   airline-alibi-index--tuna-maximum-adam
-   $ flux job cancel airline-alibi-index--tuna-maximum-adam
+  $ flux submit sleep 3600 | flux job id --to=words
+  airline-alibi-index--tuna-maximum-adam
+  $ flux job cancel airline-alibi-index--tuna-maximum-adam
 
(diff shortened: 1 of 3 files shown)


faqs.rst
```diff
@@ -256,20 +256,30 @@ systems, ``/tmp`` is a RAM-backed file system with limited space, and in
 some situations such as long running, high throughput workflows, Flux may
 use a lot of it.
 
+When the FLux database fills up the disk, errors like the following may
```
vsoch, line 259:
```suggestion
When the Flux database fills up the disk, errors like the following may
```

faqs.rst
```diff
@@ -256,20 +256,30 @@ systems, ``/tmp`` is a RAM-backed file system with limited space, and in
...
+appear and the instance of Flux will get stuck or otherwise not function
+properly
+
+.. code-block:: console
+
+ content-sqlite.err[0]: store: executing stmt: database or disk is full(13)
```
vsoch, line 265: Nit: can we be consistent with spacing for code blocks? This one has indentation that is much smaller than the others.

grondo: Sorry I copied the example right before this code block which has an indent of 1. There are also code blocks in this file with indent of 2, 3, and 4. I'll tack on an extra commit to standardize on 4.

grondo:
Thanks @vsoch! I addressed your comments. I also noticed an outdated usage of `flux start -o,-S` and fixed that.

PR #317, opened 2026-04-10
docs landing page redesign

grondo:
I worked through a redesign of the Flux landing page with Claude working as a Ruthless Technical Editor and Opinionated UX Engineer and this is the result.

The landing page is greatly simplified with a pared down "hero paragraph" and top-level links organized into cards for 3 types of users: new users (New to Flux), existing users (Using Flux), and Developers/Admins (Administration/Development).

The rest of the page is just some useful information if readers make it this far, and links off to other parts of the docs or other projects.

I've iterated a bit to get this into reasonable shape, but haven't fully checked all links for validity so there may be a lot more work here, thus I've set this as a WIP for now. If the readthedocs build works, then we'll be able to see the result via the CI build for more feedback.


garlick:
Great improvement!

faqs.rst
```diff
@@ -43,7 +42,7 @@ How does Flux compare to other managers?
 ========================================
 
 Flux has similarities to many other resource managers, but we do a variety of things quite differently.
-Check out this comparison table here (:ref:`comparison-table`) for a detailed list!"
+Check out the :ref:`comparison table <comparison-table>` for a detailed list.
```
cmoussa1, line 45: This could (and probably is) unrelated to this PR, but when the page containing the comparison table is loaded in dark mode, the **License** row still shows the license names in black text, which makes it pretty difficult to read. Maybe we should flip the font on those license to white when the page is in dark mode.

index.rst
```diff
@@ -5,66 +5,132 @@
...
+:ref:`sched:man-pages` (scheduling) |
+:ref:`security:man-pages` (security tools)
 
+---------------
+Installing Flux
+---------------
 
-Manual Pages
-------------
+**For evaluation:** `flux-core <https://github.com/flux-framework/flux-core>`_
```
cmoussa1: Probably just a me-problem, but I'm not sure I quite understand the "for evaluation" prefix here for the link to the flux-core repo. Does "evaluation" in this context mean if the user/admin is interested in actually looking at source code for flux-core? If so, that probably makes sense to me 👍 I'm honestly not even sure what I would suggest otherwise.

cmoussa1 (commented):
This is a great improvement! FWIW, I went to the ReadTheDocs PR build and checked all of the hyperlinks and they seemed to work. I had a couple of super minor comments:

index.rst
grondo: Great point! I think it is supposed to mean just to try things out vs running in production. I'm sure we can come up with something better here. The idea with these links was to give new users an idea of the basic components and their use cases.

cmoussa1: That makes total sense to me! Thanks for explaining that. I'm going to hit approve!

cmoussa1 (approved):
LGTM! Maybe we should open a separate issue on fixing the font of the licenses when viewing the comparison table in dark mode.

grondo:
BTW, I should probably mention as people notice issues please feel free to directly push them directly to this branch.

garlick (approved):
LGTM!  My vote is to move forward without further ado.

grondo:
I've pushed a couple more minor updates here. If the changes are acceptable, I can squash one of the commits down.
Summary:
- Landing page: fixed the "evaluation/production" terminology. Removed a redundant link to comparison table
- Tutorials: remove redundant filler text
- Jobs: Added formal definition of a flux job and formalized tone
- Guides: update tone to match rest of documents
- Projects: added descriptions to all core projects, added flux-pmix section
- Fixed flux-security description
- Updated flux-pmix description

cmoussa1 (approved):
This LGTM! IMHO, a very nice improvement to these index pages.

grondo:
Great! I'll squash the incremental changes commit and then set MWP. At some point we should make another pass because there's probably some other low hanging improvements we could make here.

wihobbs:
Wow! We really needed this; thanks @grondo!

grondo:
Ok, merged! Everyone please check out the new site for further improvements, broken links, etc.

PR #319, opened 2026-04-15
templates: add cross-project search using ReadTheDocs API

grondo:
Problem: The sphinx_immaterial theme uses client-side search that only searches the local project's searchindex.js file. Even though flux-docs and its subprojects (flux-core, flux-sched, flux-security, flux-rfc) are configured in ReadTheDocs as a project family, users cannot search across all documentation from the main site.

Solution: Override sphinx_immaterial's partials/search.html template to use ReadTheDocs' server-side search API with the subprojects=true parameter when on RTD (falls back to standard theme search for local builds). Add on_rtd to html_context in conf.py so it's available in templates. The custom search queries all configured subprojects and displays results with project labels (e.g., [core], [sched]).

The only way to check if this works is to let RTD build this PR and check if search is working in the preview.

### DIFF
--- a/_templates/partials/search.html
+++ b/_templates/partials/search.html
@@ -1,0 +1,185 @@
+{#- Custom search that uses ReadTheDocs server-side API for cross-project searching -#}
+{% if on_rtd %}
+  {# On ReadTheDocs, use server-side search API which supports cross-project search #}
+  <div class="md-search" data-md-component="search" role="dialog">
+    <label class="md-search__overlay" for="__search"></label>
+    <div class="md-search__inner" role="search">
+      <form class="md-search__form" name="search" id="rtd-search-form">
+        <input type="text" name="q" class="md-search__input" placeholder="Search across all Flux docs" aria-label="Search" autocapitalize="off" autocorrect="off" autocomplete="off" spellcheck="false" data-md-component="search-query" required>
+        <label class="md-search__icon md-icon" for="__search">
+          {% set icon = config.theme.icon.search or "material/magnify" %}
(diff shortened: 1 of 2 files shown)


grondo:
Ok, that didn't quite work. I'll be iterating here until it works.

grondo:
Ok, the changes "work" but we can't test the cross-site search functionality until this is merged.

garlick (approved):
Well let's give it a try then!

grondo:
Didn't seem to work 😞 I'll keep poking at it. It seems to use the right request to readthedocs so maybe something is not configured correctly there.

vsoch:
It is the Django REST Framework. So you can likely just use: https://app.readthedocs.org/api/v3/

I'm not sure we can assume that the API is being served from the node where the static site lives (did an LLM suggest that?) It's not a design that makes a lot of sense, unless readthedocs is a monolithic architecture. My suggestion is to use the above to get the exposed API. It will work for development too.

And you can find the source code here: https://github.com/readthedocs/readthedocs.org/blob/603b87bbd69071a8ad80b16c9f09543da24b6ff0/readthedocs/search/api/v3/views.py#L168

Django (and Django REST) is really great! It was always my choice for APIs.

vsoch:
But hold the phone - are you sure the "not working" isn't a lab thing? It looks like it's working here! :tada: 

image

Landing page looks great, by the way!

grondo:
Oops sorry @vsoch I missed these comments! We switched to the strategy in #320 which seems to be working.

PR #320, opened 2026-04-16
docs: use readthedocs search addons for cross-project search in main search bar

### LINKED ISSUE #318: Make the search bar search the manpages
There've been a few people (specifically Loïc, who had a question about dependencies) who probably would have benefited greatly if the search bar on Flux Framework's RTD was also able to index and search the manpages of the Framework projects. I looked into this a while back and learned a little bit about the way Sphinx is doing this; apparently, it even works offline, so it's not making any external API call (like to DuckDuckGo) we could override. More thought required...

grondo:
Supposedly if we stop trying to implement a special search box ourselves, the RTD search "Addons" feature will override the site search with its own search box which supports searching all subprojects.

Again, to test I need to create a PR so a preview will be built.

Fixes #318

### DIFF
--- a/_static/custom.css
+++ b/_static/custom.css
@@ -29,9 +29,10 @@
 }
 
 /* Sync RTD Addons search modal colors with sphinx_immaterial's slate (dark) scheme.
-   RTD Addons modal defaults to light mode; these overrides make it follow the
-   site's dark mode toggle by mapping to sphinx_immaterial's own CSS variables. */
-body[data-md-color-scheme="slate"] {
+   Variables must be set directly on the readthedocs-search element, not on body —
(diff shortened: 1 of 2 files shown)


garlick:
Ooh that might be working!

grondo:
Yeah, finally getting somewhere 🙄

grondo:
Sorry, I'm pushing incremental work here because the only way to test is the readthedocs PR builds...

grondo:
Ok, this seems to be working now. The cross-project search seems to be working in the RTD PR preview as well.

wihobbs (approved):
LGTM, thanks!

